Issue #31, opened 2014-12-29
Are bounds ignored?


kdavies4:
It seems that SALib.sample.morris_oat ignores the bounds in the parameters file.

In _params.txt_ I have:
dy.adCO.k 500 800 # mA/cm^2
dy.oxCO.k 2e-4 5e-4 # mA/cm^2

I ran `python -m SALib.sample.morris_oat -n 3 -p params.txt -o my_samples_file.txt`.

I got this in _my_samples_file.txt_:
8.88888889e-01 1.11111111e-01
3.33333333e-01 1.11111111e-01
3.33333333e-01 6.66666667e-01
8.88888889e-01 8.88888889e-01
3.33333333e-01 8.88888889e-01
3.33333333e-01 3.33333333e-01
1.00000000e+00 7.77777778e-01
1.00000000e+00 2.22222222e-01
4.44444444e-01 2.22222222e-01

None of the settings fall within the bounds of the parameters. If I change the bounds to 0.0 and 1.0, I get the same output.

Issue #28, opened 2014-12-09
Bug: Morris sample fails without error if grid_step >= num_levels


willu47:
More annoying that serious and an easy fix.  Reminder to myself to push a bug fix and write a test to solve this in forthcoming pull request.

```python

IndexError                                Traceback (most recent call last)
/site-packages/SALib/sample/morris.pyc in __init__(self, parameter_file, samples, num_levels, grid_jump, group_file, optimal_trajectories)
     62         if self.groups is None:
     63 
---> 64             self.create_sample()
     65 
     66         else:

/site-packages/SALib/sample/morris.pyc in create_sample(self)
    117                                                self.parameter_file,
    118                                                self.num_levels,
--> 119                                                self.grid_jump)
    120 
    121         else:

/Users/will2/Library/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/SALib/sample/morris_oat.pyc in sample(N, param_file, num_levels, grid_jump)
     38         for i in range(D):
     39             x_base[:, i] = (
---> 40                 rd.choice(np.arange(num_levels - grid_jump))) / (num_levels - 1)
     41 
     42         # Indices to be assigned to X, corresponding to this trajectory

/Applications/Canopy.app/appdata/canopy-1.5.1.2730.macosx-x86_64/Canopy.app/Contents/lib/python2.7/random.pyc in choice(self, seq)
    271     def choice(self, seq):
    272         """Choose a random element from a non-empty sequence."""
--> 273         return seq[int(self.random() * len(seq))]  # raises IndexError if seq is empty
    274 
    275     def shuffle(self, x, random=None):
```

jdherman:
Added by #33.

PR #33, opened 2015-01-16
Morris cleanup

jdherman:
Hi @willu47 -- 

I took a shot at merging the different Morris methods into the main class. This will hopefully reduce confusion and eliminate some duplicate code. I think a lot of the Morris-related functions don't need to be in the main class, so I put them in a separate utility file and only brought in the most important group and optimal sampling functions. Do you think this makes sense?

I also made some small changes to `read_param_file` -- the detection of groups should be automatic now without needing to toggle it anywhere. Also made the "scaled" samples the default behavior since there was some issue about that in #31. 

The test coverage is about the same as before. I'm stuck on how to test some of the sampling functions now that they're inside the Morris class. I added a test for #28 . Someday I'll get around to writing tests (at least regression ones) for the other methods. I also need to modify the other methods to use the `Sample` class.

Anyway, take a look when you have time and see what you think. This shouldn't break anything too much, it's more the organization I'm asking about. Thanks!


jdherman:
Added a little more to this, with two main changes:
- Change the Morris sampling back from a class to just a set of functions. I think this makes the API easier to use, if we can just return numpy arrays instead of saving state behind the scenes. (for example #35)

- We can have the sampling methods accept a dictionary called `problem` instead of the parameter file name. `problem` can be created either by calling `read_param_file()`, or manually like this:
```python
problem = {
   'num_vars': 3, 
   'names': ['x1', 'x2', 'x3'], 
   'groups': None, 
   'bounds': [[-3.14159265359, 3.14159265359], 
              [-3.14159265359, 3.14159265359], 
              [-3.14159265359, 3.14159265359]]
}
```
This is in response to a few requests to be able to create samples without a parameter file, which I think is a good idea. Right now only Morris is set up to do this, but I will update the other methods too.

I updated the tests to match, but I think something is weird with Travis+conda:
```
$ source activate testenv
We're making conda environments even better!  Part of this process
is changing the way environment activation works.  You need to
install the new conda-env package before you can use these feature:
    conda install -c conda conda-env
For more information, please visit: https://github.com/conda/conda-env
The command "source activate testenv" failed and exited with 1 during .
```
nosetests returns OK on my setup.

willu47:
Hi @jdherman.  Sorry about the delay in replying - I've been distracted with other work recently.  Your changes seem fair enough.  I guess the partial move to OOP wasn't fully justified and unless a wholesale effort is made to refactor the entire codebase, it is much clearer to users to keep the functional programming style.

I'll have a look at the conda issues next week after a few deadlines have passed, and push to this branch (if that's okay).

Will

jdherman:
Hi @willu47 -- no worries, I've been busy with other things here too. I think I solved the conda issue, it was just a matter of adding another line to install `conda-env` in `travis.yml`. 

Yes, that's what I was thinking about OOP. It makes everything a more organized, but also a little harder to document and test (at least without a full refactor, like you said). Another issue is that since the methods can be used as a decoupled set of commands, it's not clear if the problem properties belong to the sampling method, the analysis method, or both. This "problem dictionary" idea is a sort of compromise. 

Thanks for checking it out. I'll go ahead and merge after cleaning up a few things -- and then I'm going to finally get to writing some tests for the other methods besides morris.

willu47:
Ace!

First stop for tests could be to write high level regression tests so we can see whether the whole code package is working.  For example, it might be nice to automate the use of the Sobol G-function and write tests that call Sobol-G with different numbers of variables and see whether they correctly estimate the sensitivity indices (within an acceptable degree of error).

PR #36, opened 2015-01-29
Update regression tests

jdherman:
The coverage increases a ton, but these are only regression tests to make sure the output of the methods is about what we expect. The tests don't deal with individual functions at this point.


willu47:
Awesome!

jdherman:
The joy of regression tests...

PR #77, opened 2015-09-17
Merge Sphinx/ReadTheDocs documentation into SALib

dhadka:
Adds initial Sphinx/ReadTheDocs documentation for SALib.  Currently covers installation and basic usage.

Also added docstrings to the sample/analyze methods.  These docstrings are using numpy's docstring
format, which seems to get formatted better using Sphinx.


docs/conf.py
```diff
@@ -0,0 +1,308 @@
...
+project = u'SALib'
+copyright = u'2015, Jon Herman, Will Usher, and others'
+author = u'Jon Herman, Will Ushers, and others'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = '1.0'
```
willu47: You can add a call to the version management module `versioneer.version()` here.  You'll need to add an `import versioneer` line too.

docs/getting-started.rst
```diff
@@ -0,0 +1,74 @@
...
+    
+To install the latest development version of SALib, run the following
+commands.  Note that the development version may be unstable and include bugs.
+We encourage users use the latest stable version.
+
+::
+
+    git clone https://github.com/SALib/SALib.git
+    cd SALib
+    python setup.py install
```
willu47: You could change this to `python setup.py develop` which installs a smiling to the github repository instead of a hard install (which would be out of date as soon as you changed anything.

docs/getting-started.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Testing Installation
+--------------------
+
+To test your installation of SALib, start a new interactive Python session
+and copy/paste the code below.
+
```
willu47: You can also run the SALib tests from the command line:
    python setup.py test

docs/getting-started.rst
```diff
@@ -0,0 +1,74 @@
...
+Installing Prerequisite Software
+--------------------------------
+
+SALib requires `NumPy <http://www.numpy.org/>`_, `SciPy <http://www.scipy.org/>`_,
+and `matplotlib <http://matplotlib.org/>`_ installed on your computer.  Using
+`pip <https://pip.pypa.io/en/stable/installing/>`_, these libraries can be
+installed with:
+
+::
+
```
willu47: These three steps are not necessary, as the command `pip install SALib` automatically installs the dependencies too!

jdherman:
Just saw all of the docstrings in the sample/analyze methods, they're great. The readthedocs page looks perfect, just enough detail without being overwhelming.

My only suggestion would be to add a link to readthedocs near the top of readme.md, near all of the CI buttons and whatnot. (Or I can do that after merging).

Thanks again Dave.

willu47:
Hi @dhadka .  Thanks again for doing the grunt work and setting up the documentation.  This is really brilliant.  

One small request - please could you rename you branch from master to something else, such as "docs"?  The problem is that if I try to checkout your modifications locally (e.g. I'm trying to work out why the test failed on Travis CI), your new branch overwrites my master branch!  Generally, you should never use a master branch in a pull request because of this.

dhadka:
@willu47 Yes, I'll put your requested changes in tomorrow.  Thanks for the feedback!  The one Travis CI fail looks like it occurred during the installation of matplotlib.

dhadka:
Thanks @willu47, I will make a new docs branch to continue working on the documentation in the near term.

willu47:
No problem.  Thanks so much for doing all the hard work in getting the documentation into place.  I'll add a doc-string for the fractional factorial module too.

Is there anything we need to do to setup ReadTheDocs to work with the SALib/SALib repository?

PR #94, opened 2016-01-27
Group Sampling for Sobol Analysis and Conversion to Non-Uniform Distributions

calvinwhealton:
The changes are in two separate areas: group sampling for Sobol sensitivity analysis and conversion of uniform 0-to-1 uniform variables into other distributions. More detailed descriptions are provided below. 

Group sampling for Sobol indices makes changes to the files saltelli.py and sobol.py. In saltelli.py, the sample function was modified to add a condition for if group sampling is desired (default groups=False). If this is the case the code uses the previous method. If group sampling is desired, then instead of looping over parameters when cross-sampling the matrices, groups are looped-over with a nested loop to account for the variables within each group. The size of the matrices also changes because instead of 'D' variables in the previous code there are now 'Dg' groups. The change in sobol.py  is in the analyze function, where the additional 'groups' parameter was added and the the number of rows in the cross-sampled matrix was changed based on whether there was group sampling.

Conversion to non-uniform distributions caused changes in __init__.py and and saltelli.py. The changes in __init__py were to the scaled_samples function, where a parameter was added for 'dists' (default 'dists=None'). With the default value, the standard uniforms are converted to non-standard uniforms as before. However, when a vector of distributions is passed to the function the variables in the 'bounds' file will be used to convert the standard uniform values to the required distribution using the inverse cumulative distribution function (spicy.stats library). The possible distributions include uniform, normal, lognormal, and triangular. The change to saltelli.py is in the sample function, where an if/else condition near the end controls the call to scaled_samples and there is an additional parameter added for non-uniform distributions.

### DIFF
--- a/SALib/analyze/sobol.py
+++ b/SALib/analyze/sobol.py
@@ -19,14 +19,14 @@ except ImportError:
 
 def analyze(problem, Y, calc_second_order=True, num_resamples=100,
             conf_level=0.95, print_to_console=False, parallel=False,
-            n_processors=None):
+            n_processors=None, groups = False):
     """Perform Sobol Analysis on model outputs.
-    
+
     Returns a dictionary with keys 'S1', 'S1_conf', 'ST', and 'ST_conf', where
     each entry is a list of size D (the number of parameters) containing the
     indices in the same order as the parameter file.  If calc_second_order is
     True, the dictionary also contains keys 'S2' and 'S2_conf'.
-    
+
     Parameters
     ----------
     problem : dict
@@ -41,7 +41,9 @@ def analyze(problem, Y, calc_second_order=True, num_resamples=100,
         The confidence interval level (default 0.95)
     print_to_console : bool
         Print results directly to console (default False)
-        
+    groups : bool
+        Calculate indices based on groups (default False)
+
     References
     ----------
     .. [1] Sobol, I. M. (2001).  "Global sensitivity indices for nonlinear
@@ -56,15 +58,18 @@ def analyze(problem, Y, calc_second_order=True, num_resamples=100,
            output.  Design and estimator for the total sensitivity index."
            Computer Physics Communications, 181(2):259-270,
            doi:10.1016/j.cpc.2009.09.018.
-           
+
     Examples
     --------
     >>> X = saltelli.sample(problem, 1000)
     >>> Y = Ishigami.evaluate(X)
     >>> Si = sobol.analyze(problem, Y, print_to_console=True)
     """
-
-    D = problem['num_vars']
+    # funding number of variables/groups for sensitivity indices
+    if groups:
+        D = len(problem['groups'])
+    else:
+        D = problem['num_vars']
 
     if calc_second_order and Y.size % (2 * D + 2) == 0:
         N = int(Y.size / (2 * D + 2))
@@ -72,7 +77,7 @@ def analyze(problem, Y, calc_second_order=True, num_resamples=100,
         N = int(Y.size / (D + 2))
     else:
         raise RuntimeError("""
-        Incorrect number of samples in model output file. 
+        Incorrect number of samples in model output file.
         Confirm that calc_second_order matches option used during sampling.""")
 
     if not 0 < conf_level < 1:
@@ -97,10 +102,10 @@ def analyze(problem, Y, calc_second_order=True, num_resamples=100,
                 for k in range(j + 1, D):
                     S['S2'][j, k] = second_order(
                         A, AB[:, j], AB[:, k], BA[:, j], B)
-                    S['S2_conf'][j, k] = Z * second_order(A[r], AB[r, j], 
+                    S['S2_conf'][j, k] = Z * second_order(A[r], AB[r, j],
                         AB[r, k], BA[r, j], B[r]).std(ddof=1)
 
-    else:           
+    else:
         tasks, n_processors = create_task_list(D, calc_second_order, n_processors)
 
         func = partial(sobol_parallel, Z, A, AB, BA, B, r)
@@ -153,7 +158,7 @@ def create_Si_dict(D, calc_second_order):
     return S
 
 
-def separate_output_values(Y, D, N, calc_second_order): 
+def separate_output_values(Y, D, N, calc_second_order):
     AB = np.empty((N, D))
     BA = np.empty((N, D)) if calc_second_order else None
     step = 2 * D + 2 if calc_second_order else D + 2
@@ -193,7 +198,7 @@ def create_task_list(D, calc_second_order, n_processors):
     # Create list with one entry (key, parameter 1, parameter 2) per sobol
     # index (+conf.). This is used to supply parallel tasks to multiprocessing.Pool
     tasks_first_order = [[d, j, None] for j in range(D) for d in ('S1', 'S1_conf', 'ST', 'ST_conf')]
-    
+
     # Add second order (+conf.) to tasks
     tasks_second_order = []
     if calc_second_order:
@@ -208,7 +213,7 @@ def create_task_list(D, calc_second_order, n_processors):
     else:
         # merges both lists alternating its elements and splits the resulting list into n_processors sublists
         tasks = np.array_split([v for v in sum(
-            zip_longest(tasks_first_order[::-1], tasks_second_order), ()) 
+            zip_longest(tasks_first_order[::-1], tasks_second_order), ())
                 if v is not None], n_processors)
 
     return tasks, n_processors
@@ -272,5 +277,5 @@ if __name__ == "__main__":
                    usecols=(args.column,))
 
     analyze(problem, Y, (args.max_order == 2),
-            num_resamples=args.resamples, print_to_console=True, 
+            num_resamples=args.resamples, print_to_console=True,
             parallel=args.parallel, n_processors=args.n_processors)
--- a/SALib/analyze/sobol.py
+++ b/SALib/analyze/sobol.py
@@ -19,7 +19,7 @@ except ImportError:
 
 def analyze(problem, Y, calc_second_order=True, num_resamples=100,
             conf_level=0.95, print_to_console=False, parallel=False,
-            n_processors=None, groups = False):
+            n_processors=None):
     """Perform Sobol Analysis on model outputs.
 
     Returns a dictionary with keys 'S1', 'S1_conf', 'ST', and 'ST_conf', where
@@ -41,8 +41,6 @@ def analyze(problem, Y, calc_second_order=True, num_resamples=100,
         The confidence interval level (default 0.95)
     print_to_console : bool
         Print results directly to console (default False)
-    groups : bool
-        Calculate indices based on groups (default False)
 
     References
     ----------
@@ -65,11 +63,12 @@ def analyze(problem, Y, calc_second_order=True, num_resamples=100,
     >>> Y = Ishigami.evaluate(X)
     >>> Si = sobol.analyze(problem, Y, print_to_console=True)
     """
-    # funding number of variables/groups for sensitivity indices
-    if groups:
-        D = len(problem['groups'])
-    else:
+    # determining if groups are defined and adjusting the number
+    # of rows in the cross-sampled matrix accordingly
+    if problem['groups'] == None:
         D = problem['num_vars']
+    else:
+        D = len(problem['groups'][1])
 
     if calc_second_order and Y.size % (2 * D + 2) == 0:
         N = int(Y.size / (2 * D + 2))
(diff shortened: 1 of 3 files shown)


jdherman:
Thanks @calvinwhealton ! These are two nice features (the groups and alternate distributions). Quite a few users have asked about both recently. 

First, could you show a "quick start" example of how to use (1) the groups, and (2) the different distributions? Along the lines of the example in the readme right now. This will help get everyone on the same page.

calvinwhealton:
Thank you @jdherman for the comments. I have included some examples of how I used group sampling and non-uniform distributions. The problem specification files used for the examples are included after the code.

The problem specification files are designed to be consistent with how groups are defined in the current Method of Morris implementation. Thus, the problem specification files are of the form given below (see full examples after the code), where there is a column each for parameter name, lower bound, upper bound, group, and distribution. When group sampling is not needed, all groups can be set to 'NA' (see example 3 and related file in the sample code).
```
P1,0.0,1.0,group1,unif
P2,2.0,3.0,group2,norm
...
```

The distributions (with assumed parameter definitions) included  are uniform (upper and lower bounds), triangular (width and location of peak), normal (mean and standard deviation), and lognormal (ln-space mean and standard deviation). Therefore, the 'bounds' in the problem specification files are now distribution parameters (they may not be bounds). The triangular distribution has an assumed lower bound of zero, so the two parameters are the width and the location of the peak (in fraction of the width of the distribution).  Example 4's file has a line  `P3,2.0,.2,group1,triang`, which would mean parameter 'P3' has a triangular distribution with a width of two and the peak at a value of 20% of the width (2*0.2 =0.4). Lower bounds for the triangular and lognormal distributions can be changed from zero by the user by adding a constant to the sampled values for the parameter before computing the model output, but there is no option to change these in the problem specification files .

```
# example code for group sampling and non-uniform distributions
from SALib.util import read_param_file
import numpy as np

##### example function with several parameters to illustrate calculations
def sampleModel(x):
    y = x[:,0]**1.5 + x[:,1] + 2*x[:,2] + x[:,3] + np.exp(0.3*x[:,4]) + x[:,5] \
        + 2*x[:,1]*x[:,4] + (x[:,0]*x[:,5])**2
    return y

##### Example 1: no distributions and no groups #####
problem_nogroup_nodist = read_param_file('test_nogroup_nodist.txt')

# generating parameter values
param_values_nogroup_nodist = saltelli.sample(problem_nogroup_nodist, 10000,calc_second_order=True)

# calculating model output
Y_nogroup_nodist = sampleModel(param_values_nogroup_nodist)

# complete sensitivity analysis
Si_nogroup_nodist = sobol.analyze(problem_nogroup_nodist,Y_nogroup_nodist,calc_second_order=True,print_to_console=True)

##### Example 2: no distributions and groups ##### 
problem_group_nodist = read_param_file('test_group_nodist.txt')

# generating parameter values
param_values_group_nodist = saltelli.sample(problem_group_nodist, 10000,calc_second_order=True)

# calculating model output
Y_group_nodist = sampleModel(param_values_group_nodist)

# complete sensitivity analysis
Si_group_nodist = sobol.analyze(problem_group_nodist,Y_group_nodist,calc_second_order=True,print_to_console=True)

##### Example 3: distributions and no groups #####
problem_nogroup_dist = read_param_file('test_nogroup_dist.txt')

# generating parameter values
param_values_nogroup_dist = saltelli.sample(problem_nogroup_dist, 10000,calc_second_order=True)

# calculating model output
Y_nogroup_dist = sampleModel(param_values_nogroup_dist)

# complete sensitivity analysis
Si_nogroup_dist = sobol.analyze(problem_nogroup_dist,Y_nogroup_dist,calc_second_order=True,print_to_console=True)

##### Example 4: distributions and groups #####
problem_group_dist = read_param_file('test_group_dist.txt')

# generating parameter values
param_values_group_dist = saltelli.sample(problem_group_dist, 10000,calc_second_order=True)

# calculating model output
Y_group_dist = sampleModel(param_values_group_dist)

# complete sensitivity analysis
Si_group_dist = sobol.analyze(problem_group_dist,Y_group_dist,calc_second_order=True,print_to_console=True)

##### Example 5: distributions and groups (second example for this case) ####
problem_group_dist2 = read_param_file('test_group_dist2.txt')

# generating parameter values
param_values_group_dist2 = saltelli.sample(problem_group_dist2, 10000,calc_second_order=True)

# calculating model output
Y_group_dist2 = sampleModel(param_values_group_dist2)

# complete sensitivity analysis
Si_group_dist2 = sobol.analyze(problem_group_dist2,Y_group_dist2,calc_second_order=True,print_to_console=True)
```

Example 1 problem definition file  `test_nogroup_nodist.txt`
```
P1,0.0,1.0
P2,0.0,1.0
P3,0.0,1.0
P4,0.0,1.0
P5,0.0,1.0
P6,0.0,1.0
```

Example 2 problem definition file  `test_group_nodist.txt`
```
P1,0.0,1.0,group1
P2,2.0,3.0,group2
P3,0.5,1.0,group1
P4,0.0,5.0,group1
P5,-.5,.5,group2
P6,0.0,1.0,group3
```

Example 3 problem definition file  `test_nogroup_dist.txt`
```
P1,0.0,1.0,NA,unif
P2,1.0,.75,NA,triang
P3,0.0,.2,NA,norm
P4,0.0,.2,NA,lognorm
P5,-1.0,1.0,NA,unif
P6,1.0,.25,NA,triang
```

Example 4 problem definition file `test_group_dis.txt`
```
P1,0.0,1.0,group1,unif
P2,1.0,.75,group2,triang
P3,2.0,.2,group1,triang
P4,0.0,.2,group1,unif
P5,0.0,1.0,group1,unif
P6,1.0,.25,group3,triang
```

Example 5 problem definition file  `test_group_dist2.txt`
```
P1,0.0,1.0,group1,unif
P2,1.0,.75,group2,triang
P3,0.0,.2,group1,norm
P4,0.0,.2,group1,lognorm
P5,-1.0,1.0,group2,unif
P6,1.0,.25,group3,triang
```

jdherman:
Thanks @calvinwhealton this is great. I'm glad you looked at how Will did the Morris groups to try to make it consistent. Would you mind a few more questions? :)

- It's great that the parameter file works. Does the problem dictionary also work? For example:
```python
problem = {
  'num_vars': 3, 
  'names': ['x1', 'x2', 'x3'], 
  'bounds': [[-np.pi, np.pi]]*3
  'groups': ['A', 'A', 'B'],
  'dists': ['unif', 'norm', 'triang']
}

param_values = saltelli.sample(problem, 1000, calc_second_order=True)
```

- Are the distributions specific to Sobol, or could they be used by any of the sampling methods? (I imagine they could). If so, do we have to change the term "bounds" to something else throughout the whole code base, if we want to technically be correct?

- It looks like there is some issue with the build related to the `scale_samples` function receiving too few arguments. Do you use the `dists` argument in `scale_samples`? 

Over the weekend I'll try to pull this down and write some tests against it. This is really useful and I want to make sure we can add all of the new features without disrupting anyone's existing code.

calvinwhealton:
Thanks @jdherman for mentioning these points. Here are my brief responses.

- The `saltelli.sample` function uses the matrix that `read_param_file` generates and is stored as `problem['groups'][0]`. This matrix is used in the cross-sampling portion of the code because it seemed relatively clean. The method of specifying the problem that you show would not be consistent with this because the "groups" entry in the dictionary is not a tuple and there is no matrix. It is possible to add a separate condition based on the structure "groups" to allow problem definition as shown in your example.

- The distributions should be able to be used with other methods besides Sobol. To be technically correct I think that changing "bounds" to something else would be better, but I preserved the existing naming convention to avoid causing problems throughout the library. Alternatively, "bounds" could be preserved and taken to represent the central 95% interval for the normal and lognormal distributions. With this information the distribution parameters could be calculated based on this information, although some generated values would be outside the "bounds". The asymmetric triangular distribution still have problems being defined in terms of "bounds". I welcome other suggestions on how to represent the distribution parameters in a way that will be helpful to users.

- You are correct that there was a third parameter that I forgot to remove after splitting the `nonuniform_scale_samples` and the `scale_samples` functions. This has been removed in the latest commit and the examples I posted still ran on my machine after this change.

I would appreciate any comments on how to make the code more consistent with existing portions of the library or more intuitive to the users.

jdherman:
Ok got it, thanks. I can play around with the "groups" thing after we merge. I'll also try to write some unit tests. I don't think we should change the name "bounds" right now .... especially because the proper name would be "parameters", and that will cause all kinds of confusion with the actual "model parameters" vs. the parameters of the distributions ..... etc etc.

Before we merge we'll have to have the Travis CI tests passing (see below). Right now the issue is that wherever these lines show up:
```python
if problem['groups'] == None:
```
or
```python
if problem['dists'] == None:
```
It throws an error if those keys don't exist. Sometimes if people aren't using parameter files, they won't have those keys defined in the problem dictionary. So, the best way around this that I know of would be to check:
```python
if not problem.get('groups'):
```
The `get` function for dictionaries will automatically return `None` if the key doesn't exist, rather than throwing an error for key not found.

We'll just keep cracking away at these things until all of the tests pass!

calvinwhealton:
Thanks for the suggestion on reformulating the if conditions. Also, I added the capability for the user to define the problem dictionary in the code without having to read-in the parameter file. Some examples are given below, corresponding to the examples I provided in a previous post.

```
# example code for group sampling and non-uniform distributions
from SALib.util import read_param_file
import numpy as np

##### example problem with several parameters to illustrate calculations
def sampleModel(x):
    y = x[:,0]**1.5 + x[:,1] + 2*x[:,2] + x[:,3] + np.exp(0.3*x[:,4]) + x[:,5] \
        + 2*x[:,1]*x[:,4] + (x[:,0]*x[:,5])**2
    return y

##### Example 1: no distributions and no groups #####
problem_nogroup_nodistB = {
    'num_vars':6,
    'names': ['P1','P2','P3','P4','P5','P6'],
    'bounds':[[0.0,1.0]]*6
}

# generating parameter values
param_values_nogroup_nodistB = saltelli.sample(problem_nogroup_nodistB, 10000,calc_second_order=True)

# calculating model output
Y_nogroup_nodistB = sampleModel(param_values_nogroup_nodistB)

# complete sensitivity analysis
Si_nogroup_nodistB = sobol.analyze(problem_nogroup_nodistB,Y_nogroup_nodistB,calc_second_order=True,print_to_console=True)

##### Example 2: no distributions and groups ##### 
problem_group_nodistB = {
    'num_vars':6,
    'names': ['P1','P2','P3','P4','P5','P6'],
    'bounds':[[0.0, 1.0], [2.0, 3.0], [0.5, 1.0], [0.0, 5.0], [-0.5, 0.5], [0.0, 1.0]],
    'groups':['group1','group2','group1','group1','group2','group3']
}

# generating parameter values
param_values_group_nodistB = saltelli.sample(problem_group_nodistB, 10000,calc_second_order=True)

# calculating model output
Y_group_nodistB = sampleModel(param_values_group_nodistB)

# complete sensitivity analysis
Si_group_nodistB = sobol.analyze(problem_group_nodistB,Y_group_nodistB,calc_second_order=True,print_to_console=True)

##### Example 3: distributions and no groups #####
problem_nogroup_distB = {
    'num_vars':6,
    'names': ['P1','P2','P3','P4','P5','P6'],
    'bounds':[[0.0,1.0], [1.0, 0.75], [0.0, 0.2], [0.0, 0.2], [-1.0,1.0], [1.0, 0.25]],
    'dists':['unif','triang','norm','lognorm','unif','triang']
}

# generating parameter values
param_values_nogroup_distB = saltelli.sample(problem_nogroup_distB, 10000,calc_second_order=True)

# calculating model output
Y_nogroup_distB = sampleModel(param_values_nogroup_distB)

# complete sensitivity analysis
Si_nogroup_distB = sobol.analyze(problem_nogroup_distB,Y_nogroup_distB,calc_second_order=True,print_to_console=True)

##### Example 4: distributions and groups #####
problem_group_distB = {
    'num_vars':6,
    'names': ['P1','P2','P3','P4','P5','P6'],
    'bounds':[[0.0,1.0], [1.0, 0.75], [2.0, 0.2], [0.0, 0.2], [0.0,1.0], [1.0, 0.25]],
    'dists':['unif','triang','triang','unif','unif','triang'],
    'groups':['group1','group2','group1','group1','group2','group3']
}

# generating parameter values
param_values_group_distB = saltelli.sample(problem_group_distB, 10000,calc_second_order=True)

# calculating model output
Y_group_distB = sampleModel(param_values_group_distB)

# complete sensitivity analysis
Si_group_distB = sobol.analyze(problem_group_distB,Y_group_distB,calc_second_order=True,print_to_console=True)

##### Example 5: distributions and groups (second example for this case) ####
problem_group_dist2B = {
    'num_vars':6,
    'names': ['P1','P2','P3','P4','P5','P6'],
    'bounds':[[0.0,1.0], [1.0, 0.75], [0, 0.2], [0.0, 0.2], [-1.0,1.0], [1.0, 0.25]],
    'dists':['unif','triang','norm','lognorm','unif','triang'],
    'groups':['group1','group2','group1','group1','group2','group3']
}

# generating parameter values
param_values_group_dist2B = saltelli.sample(problem_group_dist2B, 10000,calc_second_order=True)

# calculating model output
Y_group_dist2B = sampleModel(param_values_group_dist2B)

# complete sensitivity analysis
Si_group_dist2B = sobol.analyze(problem_group_dist2B,Y_group_dist2B,calc_second_order=True,print_to_console=True)
```

PR #95, opened 2016-02-07
Sobol groups and distributions

jdherman:
Incorporates all updates from #94 and some tests.

@calvinwhealton check it out and see what you think. As far as I can tell, it still does all the same stuff. I really appreciate the additions.

@willu47 please take a look at this too, because I changed the way that read_param_file creates groups. But only internally, I don't think any external API should change. Generating the matrix is now done separately. The goal for this was to make sure that read_param_file will return a dictionary that's exactly the same format as the "problem dictionary" that someone would use instead of the param file.

Once we get this merged, there will be a good chance to get nonuniform distributions set up for all of the methods. 

Thanks!

### DIFF
--- a/SALib/analyze/morris.py
+++ b/SALib/analyze/morris.py
@@ -6,7 +6,7 @@ from scipy.stats import norm
 import numpy as np
 
 from . import common_args
-from ..util import read_param_file
+from ..util import read_param_file, compute_groups_matrix
 
 
 def analyze(problem, X, Y,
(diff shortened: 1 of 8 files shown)


calvinwhealton:
@jdherman the code looks good to me. Thanks for improving my implementation.

jdherman:
@willu47 let us know if you have any thoughts on the group sampling before we merge this. The API should work the same, but internally the group matrix is handled a little bit differently, so just wanted to check your opinion.

willu47:
Hi @jdherman, sorry for the tardy response.  I'm completely happy with the new groups paradigm.  As long as the tests pass, and the API remains the same, what goes on behind the scenes can change as required.  I've made some comments inline.

jdherman:
Thanks Will. Small updates to fix the points you caught, and another indexing bug that showed up during testing. 

I think this is good to go. If it passes Travis I'll merge, bump the version by some negligible increment, and push to PyPi. 

Thanks Calvin! Now you're responsible for a blog post on WaterProgramming about how to use these new features :)

PR #97, opened 2016-02-25
fix dists without groups issue

### LINKED ISSUE #96: dists without groups causes error
from @calvinwhealton ...

It seems that reading-in a problem file with no groups but with distributions is triggering an error in lines 183-185 of __init__.py. The error is that there are no groups so the results will not be meaningful and it does not read-in the file. Was this an intentional change because of some of the methods (Morris)? I think a simple work around is to define groups and once the problem file is read-in to set the ‘groups' to None, although it is inelegant.

jdherman:
fixes #96


jdherman:
Something broke on the Travis side, will try again later.

PR #105, opened 2016-10-11
Added paper and bibiography for submission to JOSS

willu47:


jdherman:
This is great, thanks Will.

Issue #109, opened 2016-10-25
Odd Sobol first order indices


Anto-F:
Hello,
I compare some tools and methods to calculate Sobol indices and I got some strange values for first order indices with SALib.

With Polynomial Chaos (openturns) and EASI algorithm, I have the following results for S1's:

```
    Polynomial Chaos    EASI
x1  0.48                0.48
x2  0.25                0.25
x3  0.10                0.11
x4  0.04                0.05
x5  0.04                0.04
x6  0.03                0.03
x7  0.02                0.02
x8  0.02                0.02
x9  0.02                0.02
x10 0.01                0.01
```

As you can see, they converge to the same values.
And 500 simulations with a Quasi Monte-Carlo sample is enough to reach these values (I tried up to 2000 simulations and they remain the same).
Also, the ST's I got with Polynomial Chaos are similar to the S1's (EASI doesn't include this calculation).

With Salib I use saltelli.sample to generate 5000 samples and I have the following results:

```
    S1      S1_conf  ST  ST_conf
x1  0.92    4.99    0.53    0.06
x2  0.29    2.37    0.16    0.02
x3  0.22    1.70    0.11    0.01
x4  0.25    1.24    0.05    0.01
x5  0.04    1.14    0.04    0.01
x6  0.03    0.98    0.03    0.00
x7  -0.09   1.01    0.02    0.00
x8  0.05    0.98    0.02    0.00
x9  0.06    0.83    0.02    0.00
x10 0.02    0.82    0.01    0.00
```

Except for the second parameter, the ST's are really close to what I got with other methods.
However, the S1's don't look good and the confidence intervals are terrible.
I don't understand why SALib is able to calculate ST's but not S1's in my case, specially if they have similar values.
I could increase the sample size but 5000 took 80 hours...
Any thoughts on this issue ?

jdherman:
Thanks @Anto-F . This sounds similar to #102 . At first I thought this was a function-specific problem but now I'm wondering if there is a bug in the S1 calculation.

I'll look into it. Could you say more about the function?

Anto-F:
I read the #102 before posting and yes, it sounds similar.
My function is a whole software, [EnergyPlus](http://apps1.eere.energy.gov/buildings/energyplus/), for building energy simulation.

jdherman:
Ok. A couple of questions:
- Is the model deterministic? 
- Do you expect a high degree of interactions?
- Can you confirm that the Ishigami function example in the readme works?

[...]

PR #128, opened 2017-01-07
Bugfix 114 and 127

### LINKED ISSUE #114: Centering output for sobol.analyze
Hi,
I have a question regarding sobol.analyze: how come there is no option for automatically normalizing/centering the output? I found the following warning for the method "sobol2002" in the R-package sensitivity (https://cran.r-project.org/web/packages/sensitivity/sensitivity.pdf):

'BE CAREFUL! This estimator suffers from a conditioning problem when estimating the variances
behind the indices computations. This can seriously affect the Sobol’ indices estimates in case
of largely non-centered output. To avoid this effect, you have to center the model output before
applying "sobol2002". ...'

Would this apply also for sobol.analyze? And in addition, could this explain the issue #109?

### LINKED ISSUE #127: morris analysis method requires problem['group'] to exist
The sample method uses problem.get('groups') and the analysis method should, too, to avoid requiring call to set problem['group'] = None.

jdherman:
A couple of small changes to clean up before we release an updated version.

* Fixes #127 (using `problem.get('groups')` in analyze.morris)

* Fixes #114, and by extension #109, by normalizing the model output in analyze.sobol using the suggestion by @sofianehaddad (`Y = (Y - Y.mean()) / Y.std()`)

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2013-2015 Jon Herman, Will Usher, and others.
+Copyright (c) 2013-2017 Jon Herman, Will Usher, and others.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


jdherman:
@willu47 this is all small stuff so I'll go ahead and merge

PR #138, opened 2017-03-31
fix for sobol sequence: first row should be zero not empty. add test …

### LINKED ISSUE #136: First row of sobol_sequence()
Hi,

I'm trying to use your sample.sobol_sequence function but I've noticed that the first row is giving completely wrong values (like 2.13393829e+219). It seems that you are skipping first rows in your code when you use sobol_sequence() but I'm not sure to understand what is happening here.
Is it an expected behavior or am I doing something wrong?

Thank you for your work !

jdherman:
…against Joe&Kuo example

Looking at the original C++ code from Joe and Kuo (2008):
http://web.maths.unsw.edu.au/~fkuo/sobol/

The first row of the Sobol sequence matrix should be set to zero, as it is never assigned anywhere else in the function.

Fixes #136

### DIFF
--- a/SALib/analyze/delta.py
+++ b/SALib/analyze/delta.py
@@ -96,7 +96,7 @@ def calc_delta(Y, Ygrid, X, m):
 # Plischke et al. 2013 bias reduction technique (eqn 30)
 
 def bias_reduced_delta(Y, Ygrid, X, m, num_resamples, conf_level):
-    d = np.empty(num_resamples)
+    d = np.zeros(num_resamples)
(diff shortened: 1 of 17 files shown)


willu47:
👍

willu47:
`np.empty` has tripped me up quite a few times before.  It might be worth doing a find-replace throughout the project to `np.zeros` (which is usually what is required).

jdherman:
Yes, down with `np.empty`! Thanks Will.

PR #147, opened 2017-06-23
continuous deployment when tags pushed

jdherman:
As suggested by @willu47 in #142 

Have not tested this, because it will only deploy from master. We'll have to merge and then test it by pushing some tags:
```
git tag v1.0.x
git push --tags
```
(at least this is how I've been doing it -- there is probably a better way)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
-# Travis config for SALib/SALib
 sudo: false
 language: python
 virtualenv:
   system_site_packages: true
 services:
-  - postgresql
+- postgresql
 env:
   matrix:
   - DISTRIB="conda" PYTHON_VERSION="2.7" COVERAGE="true"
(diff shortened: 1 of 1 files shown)


willu47:
Looks great @jdherman.  Here's some notes from another repo where we set this up.

To make a release, create an annotated tag, and submit a pull request:
```bash
git tag -a v0.2.0 -m "A release message"  # create annotated tag with message
git describe            # show current commit in relation to tags
```
You’ll need to specify you tag to push either using the --tags flag or the tag name:
```bash
git push upstream master --tags
git push upstream v0.2.0        # alternatively
```

jdherman:
Thanks @willu47. I pushed a new tag (v1.0.4) about 20 minutes ago but I'm not seeing anything updated on pypi. Is there a way to debug this / how long does it usually take?

willu47:
Hi @jdherman.  I created an *annotated* tag, and force-pushed it.  This has kicked off a new build on travis.  At the bottom of the build step (after the tests have run), you should be able to see the results of the deploy step.

Issue #155, opened 2017-07-27
Confusing kwarg values in the documentation


Armavica:
The guide, on [this page](http://salib.readthedocs.io/en/latest/basics.html) for example, mentions two uses of kwargs:

```python
param_values = saltelli.sample(problem, 1000, calc_second_order=True)
```
and
```python
Si = sobol.analyze(problem, Y, print_to_console=False)
```
that I find confusing. Seeing that these options are explicitly set, I am guessing that this was to override the default behavior, hence if I want the opposite behavior, I can just remove them. But this is incorrect since these values are in fact redundant with the default behavior.

If I were you, I would not set these parameters explicitly in this example since this is redundant and can be confusing. If the purpose was to insist on these options at this point in the documentation, then I would use them with the opposite (non-default) behavior.

PR #156, opened 2017-07-27
do not use default kwargs in docs and readme examples

jdherman:
Addresses #155 

In the readme I just changed them to the non-default options. This way it still explains what the argument does without being redundant.

In the docs I removed the kwargs from the function call and describe them in the text instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,13 +42,13 @@ problem = {
 }
 
 # Generate samples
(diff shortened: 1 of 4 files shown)


willu47:
This seems all good - sorry it has taken so long for me to review it!

Issue #161, opened 2017-09-13
Documentation problem['groups']


rJonatan:
Hi!
I'd like to understand better the meaning of 'groups' in the problem definition. Could someone please explain how to use it?

Currently, the API documentation only comments groups for the method of Morris (http://salib.readthedocs.io/en/latest/api.html#method-of-morris), but judging from the source code it applies also to Sobol: http://salib.readthedocs.io/en/latest/_modules/SALib/analyze/sobol.html. Not for FAST however.

willu47:
Hi @rJonatan, thanks a lot for the issue.  You are absolutely right, the documentation on groups is poor.  I implemented the groups functionality for Morris.  You can read more about the approach with groups in section 3.5 of [Saltelli 2008](http://www.andreasaltelli.eu/file/repository/A_Saltelli_Marco_Ratto_Terry_Andres_Francesca_Campolongo_Jessica_Cariboni_Debora_Gatelli_Michaela_Saisana_Stefano_Tarantola_Global_Sensitivity_Analysis_The_Primer_Wiley_Interscience_2008_.pdf).

In SALib, to define a problem file with groups, you add an extra column in which you name the group to which a parameter belongs. For example:
```
x1 -3.14 3.14 Group_1
x2 -3.14 3.14 Group_2
x3 -3.14 3.14 Group_2
```
This is loaded into SALib and creates a problem dictionary with three parameters with ranges [0, 1] and two groups.  `Group 1` contains `Param 1` and `Param 2` and `Group 2` contains `Param 3`.  The dict which is populated in SALib (and which you can pass directly to the sample and analyze methods) looks like this:
```python
{
'groups': ([0, 1, 1], ['Group_1', 'Group_2']),
'names': ['x1', 'x2', 'x3'],
'num_vars': 3,
'bounds': [[-3.14, 3.14], [-3.14, 3.14], [-3.14, 3.14]]
}
```
The `groups` value contains a tuple in which the first entry is a list which maps the parameters to groups, and the second entry contains a list of the group names. That is all there is to it.

@jdherman - I dimly remember that groups were added to Sobol, could you confirm?

jdherman:
Good find. Yes, the groups feature is added to Sobol, but not the other methods. Unfortunately I'm seeing that it's using a slightly different API than what Will mentioned above.

Here is a working Sobol example with groups:
```python
problem = {
  'groups': ['Group_1', 'Group_2', 'Group_2'],
[...]

PR #163, opened 2017-09-19
Patch docs groups

jdherman:
Addresses #161, there was some confusion in the purpose and syntax of parameter groups.

* added a section to advanced readme
* fixed docstring for the `read_param_file()` function in `util/__init__.py`
* removed redundant input parameter `num_vars` from `compute_groups_matrix` function in the same file, and changed all calls to this function throughout. (The number of variables is always the length of the list of group names, which is the other input)

No changes to dependencies, so not sure why the dependency-ci error below.

### DIFF
--- a/README-advanced.md
+++ b/README-advanced.md
@@ -1,18 +1,18 @@
-### SALib: Advanced options
+## SALib: Advanced options
 
-#### Parameter files
+### Parameter files
 
-In the parameter file, lines beginning with `#` will be treated as comments and ignored. The Morris method also supports groups of input factors, which can be specified with a fourth column:
+In the parameter file, lines beginning with `#` will be treated as comments and ignored.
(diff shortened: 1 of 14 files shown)


jdherman:
Something to note -- a few of the more advanced Morris-related methods are still using the tuple version of `groups` under the hood. So they do something like this:

```
groups = compute_groups_matrix(problem['groups'])
# `groups` is now a tuple of (group_matrix, unique_group_names)
```
As far as I can tell this is all hidden from the user.

willu47:
@jdherman - working on removing those unnecessary calls to `compute_groups_matrix` now.

jdherman:
Nice, I didn't realize those functions only needed `num_groups` and not the matrix. Very clean now.

Merging! Thanks.

Issue #162, opened 2017-09-18
Minimum number of FAST samples


Bleevoe:
It seems to me there are two bugs related to minimum number of samples for FAST.

1. I think FAST requires N > 4M^2. However, the first line in fast_sampler.sample instead checks that N >= 4M^2. So when I choose N = 4M^2, I get some NaN values.
2. If I instead choose N = 4M^2 + 1, I get an IndexError in the analyze step. I have not tried to understand why.

Below is a script for reproducing the two above problems. Choosing N >= 4M^2 + 2 seems to work fine. I am using SALib 1.0.7, Python 2.7.11, and NumPy 1.10.4 on Windows 10.

```python
from SALib.sample import fast_sampler
from SALib.analyze import fast
from SALib.test_functions import Ishigami
import numpy as np

problem = {
  'num_vars': 3,
  'names': ['x1', 'x2', 'x3'],
  'bounds': [[-np.pi, np.pi]]*3
}

# NaN with N = 64
N = 64
samples = fast_sampler.sample(problem, N)
print('NaN occurrences with N = 64: %d' % np.sum(np.isnan(samples)))

# Weird error with N = 65
N = 65
samples = fast_sampler.sample(problem, N)
Y = Ishigami.evaluate(samples)
try:
    Si = fast.analyze(problem, Y)
except IndexError as e:
    print("Weird error with N = 65: %s" % e)

# All green with N = 66
N = 66
samples = fast_sampler.sample(problem, N)
Y = Ishigami.evaluate(samples)
Si = fast.analyze(problem, Y)
```

jdherman:
@Bleevoe thanks for catching these and for the very informative test example! Sorry it took a while to get around to it.

First one is easy, I just messed up the check. Changed to:
```python
    if N <= 4*M**2:
        raise ValueError("""
        Sample size N > 4M^2 is required. M=4 by default.""")
```
Second one is trickier. I ran your code in a loop, and the error occurs for many values of N, odd numbers only:
```
ERROR ON N =  65
ERROR ON N =  73
ERROR ON N =  81
ERROR ON N =  89
ERROR ON N =  97
ERROR ON N =  105
ERROR ON N =  113
ERROR ON N =  121
ERROR ON N =  129
ERROR ON N =  137
ERROR ON N =  145
ERROR ON N =  153
ERROR ON N =  161
ERROR ON N =  169
ERROR ON N =  177
ERROR ON N =  185
ERROR ON N =  193
... (and many more)
```

[...]

PR #165, opened 2017-10-08
remove quotes and commas from setup.cfg classifiers

jdherman:
I think this will resolve the deployment error at the end of #162 ... looks like Travis does not expect quotes or commas anymore, and adds the quotes instead.

Edit: this is only a guess and might not be correct. Is there a way to see if it works without merging first?


willu47:
That should be fine.  You can type `python setup.py --classifiers` to check from the command line, and should see entries which match those in the [pypi list](https://pypi.python.org/pypi?%3Aaction=list_classifiers).

Issue #174, opened 2017-11-21
Nonuniform distributions support for method of morris sampling


filssavi:
I'd like to contribute nonuniform sampling support for the method of morris.

I have hacked together a proof of concept based on the util.nonuniform_scale_samples function used in the sobol sampling algorithm.

The only problem in an otherwise trivial implementation is that the ppf function used explodes (outputs -inf or inf respectively) whenthe inputs get too close to 1 or 0
The approach I used to overcome this I just crop the allowable sample range so that only values inside a six sigma range from the mean value are returned.

Since I'm just a user of the library and I'm not extremely familiar with this method's inner workings I open this issue to discuss the enhancement before the pull request

jdherman:
Hi @filssavi thanks for the contribution! This sounds like a good idea. Please go ahead and submit a PR. I'll also ask @calvinwhealton to join the discussion, since he implemented the nonuniform distributions for Sobol.

I looked at your branch and did not see the fix for the -inf/inf issue, was that in there too? That's also something we should have caught before.

A question to @willu47 and others. Does anyone see a case where nonuniform samples would mess up one of the methods? (Morris, FAST) For example I'm wondering if we should not include the option for Latin Hypercube, since then it would not technically be a LHS anymore.

filssavi:
yesterday i was just playing with the code in a notebook, i just pushed to my fork the code with all the support in place and the -inf/inf issue solved

I added support to the following methods:
    -morris
    -latin hypercube
    -FAST
 
I checked the resulting distributions by plotting a histogram and they seemed ok, i haven't tested them more thoroughly, since i don't have such a strong background.

not passing the dist argument to the problem dict results in standard behaviour just as before,

PR #175, opened 2017-11-22
Nonuniform sampling support

filssavi:
solution to issue #174

### DIFF
--- a/SALib/sample/ff.py
+++ b/SALib/sample/ff.py
@@ -8,7 +8,7 @@ This implementation is based on the formulation put forward in
 from scipy.linalg import hadamard
 import numpy as np
 from . import common_args
-from ..util import nonuniform_scale_samples, scale_samples, read_param_file
+from ..util import scale_samples, read_param_file
 
 
 def find_smallest(num_vars):
--- a/SALib/sample/ff.py
+++ b/SALib/sample/ff.py
@@ -8,7 +8,7 @@ This implementation is based on the formulation put forward in
 from scipy.linalg import hadamard
 import numpy as np
 from . import common_args
-from ..util import scale_samples, read_param_file
+from ..util import nonuniform_scale_samples, scale_samples, read_param_file
 
 
 def find_smallest(num_vars):
@@ -105,8 +105,15 @@ def sample(problem):
     contrast = generate_contrast(problem)
     sample = np.array((contrast + 1.) / 2, dtype=np.float)
     problem = extend_bounds(problem)
-    scale_samples(sample, problem['bounds'])
-    return sample
+
+    if not problem.get('dists'):
+        # scaling values out of 0-1 range with uniform distributions
+        scale_samples(sample, problem['bounds'])
+        return sample
+    else:
+        # scaling values to other distributions based on inverse CDFs
+        scaled_sample = nonuniform_scale_samples(sample, problem['bounds'], problem['dists'])
+        return sample
 
 if __name__ == "__main__":
 
--- a/SALib/sample/ff.py
+++ b/SALib/sample/ff.py
@@ -76,6 +76,7 @@ def generate_contrast(problem):
 
 def sample(problem):
     """Generates model inputs using a fractional factorial sample
+
     Returns a NumPy matrix containing the model inputs required for a
     fractional factorial analysis.
     The resulting matrix has D columns, where D is smallest power of 2 that is
@@ -99,8 +100,8 @@ def sample(problem):
     contrast = generate_contrast(problem)
     sample = np.array((contrast + 1.) / 2, dtype=np.float)
     problem = extend_bounds(problem)
-    scale_samples(sample, problem['bounds'])
-    return sample
+    scaled_sample = scale_samples(sample, problem)
+    return scaled_sample
 
 
 if __name__ == "__main__":
--- a/SALib/sample/ff.py
+++ b/SALib/sample/ff.py
@@ -76,44 +76,32 @@ def generate_contrast(problem):
 
 def sample(problem):
     """Generates model inputs using a fractional factorial sample
-
     Returns a NumPy matrix containing the model inputs required for a
     fractional factorial analysis.
     The resulting matrix has D columns, where D is smallest power of 2 that is
     greater than the number of parameters.
     These model inputs are intended to be used with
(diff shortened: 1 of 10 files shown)


filssavi:
Upon further testing there are few bugs still left to iron out, moreover as the distance from the bound increases a spike the edge of the limited sample's distibution is introduced, I have to sort this issue prior to merging

So I will close this pull request and make a new one once this issues are fixed

filssavi:
the last bugs have been ironed out and the problems at the bounds have been fixed by proper rescaling prior to the distribution change

jdherman:
Hi @filssavi thanks for working on this. Nonuniform sampling will be a nice addition to the other methods.

I didn't think it would need to be so complicated, though. In the `saltelli` sampling method (the one that existed before), the code is only this:
```python
    if not problem.get('dists'):
        # scaling values out of 0-1 range with uniform distributions
        scale_samples(saltelli_sequence, problem['bounds'])
        return saltelli_sequence
    else:
        # scaling values to other distributions based on inverse CDFs
        scaled_saltelli = nonuniform_scale_samples(saltelli_sequence, problem['bounds'], problem['dists'])
        return scaled_saltelli
```

I thought we could just add a similar if-then block to the other methods. In this PR it seems that things are implemented differently across methods, and I also don't understand the need for any bounds checking or cropping the range of samples.

Part of the confusion could be the way that `nonuniform_scale_samples` works, because the parameters that we call "lower and upper bounds" (for uniform distributions) are used differently for the other distributions:
```python
def nonuniform_scale_samples(params, bounds, dists):
    """Rescale samples in 0-to-1 range to other distributions

    Arguments
    ---------
    problem : dict
        problem definition including bounds
    params : numpy.ndarray
        numpy array of dimensions num_params-by-N,
        where N is the number of samples
    dists : list
        list of distributions, one for each parameter
            unif: uniform with lower and upper bounds
            triang: triangular with width (scale) and location of peak
                    location of peak is in percentage of width
                    lower bound assumed to be zero
            norm: normal distribution with mean and standard deviation
            lognorm: lognormal with ln-space mean and standard deviation
    """
```
The samples from the nonuniform distributions do not need to fall between any bounds. The [0,1] samples are put into the inverse CDF to generate samples from the chosen distribution.

So, the implementation can be simpler and cleaner across methods. I appreciate the time investment and will keep the PR open if you want to keep working on it.

Thanks,
Jon

filssavi:
The first test I did was just added the if-then-else block from saltelli sampling to the method of morris one, however (probably due to differences in the sample generating algorithms infinities started to come out

The core problem is that for infinite distributions (normal and lognormal in this case) the following are true:

codecogseqn
codecogseqn 1

so if 0 or 1 is present in the original (unscaled and unshifted) samples list, when resampling with  `nonuniform_scale_samples` -inf and +inf will be output, this while technically correct is unhelpful, since:

- most models are not built to take into account infinite inputs so they will most definitely crash or refuse to work entirely

- when using sensitivity analysis to asses the validity of model of some phenomenon happening in the real world, the values that a parameter, might follow a normal/lognormal distribution, but some nonlinear effect will limit the range of possible values, an example (from wikipedia) is  blood pressure of adult humans, however at some point blood vessels will start to rupture, so the tails of the distributions will not go up to infinity


For this reason, the values of the original samples (before any rescaling), so a reasonable assumption is that for a normal distribution a reasonable default could be to consider the six sigma range around the mean value.

However since it comes essentially for free I thought to offer users of the library the possibility of overriding this default and manually choosing the range of the normal distribution form which to sample,

in the case of the saltelli sampling method (with which i'm not too familiar so what follows it's pure speculation), I think this problems haven't shown up because the original samples are already chosen in a way that excludes the boundary values

SALib/sample/ff.py
```diff
@@ -76,30 +76,24 @@ def generate_contrast(problem):
 
 def sample(problem):
     """Generates model inputs using a fractional factorial sample
-
```
willu47: These blank lines should be retained in the docstring to comply with numpydoc style.

SALib/sample/latin.py
```diff
@@ -39,8 +39,23 @@ def sample(problem, N):
         for j in range(N):
             result[j, i] = temp[j]
 
-    scale_samples(result, problem['bounds'])
```
willu47: Could all this functionality be pushed into the `scale_samples` function if the problem dict was passed to it?  This way, there would be no duplication of the code across the different methods?

SALib/util/__init__.py
```diff
@@ -280,3 +280,70 @@ def _wrapper(*args, **kwargs):
...
+            bounds.append([lower_bound,upper_bound])
+        else:
+            bounds.append([0,1])
+
+    limited_sample = samples
+    scale_samples(samples,bounds)
+
+    return limited_sample
+
+def checkBounds(problem):
```
willu47: To comply with python style guides, camel case should not be used please use lower case and underscores for function names.

SALib/util/__init__.py
```diff
@@ -280,3 +280,70 @@ def _wrapper(*args, **kwargs):
...
+    """
+    if not problem.get('dists_upper_bound'):
+        upper_bound = 0.999999998026825
+    else:
+        upper_bound = problem.get('dists_upper_bound')
+        if not 0 < upper_bound < 1:
+            raise ValueError("Nonuniform distribution value range invalid")
+
+    if not problem.get('dists_lower_bound'):
+        lower_bound = 1 - 0.999999998026825
```
willu47: `lower_bound = 1 - upper_bound` would be neater

filssavi:
I implemented most of the required changes (tests are still missing, i'll add them when I have some time)

the only issue remaining is with the fast method, somehow the change makes the regression test fail (the analisys results are different), I checked and superficially there seems to be nothing wrong with the sampling part of the code

since I don't really know how fast inner working and all other tests are passing I suspect the sampling part is working fine and I somehow triggered a bug in the fast analize implementation, can someone confirm/deny this

thanks

willu47:
@filssavi - I'm now back in my office and can take another look at these failing tests and get this pull request merged soon.

willu47:
Closing this branch in preference to the rebased (and renamed from filssavi/master) #192.

PR #187, opened 2018-01-21
fix for 'one' instead of 'zero' in ValueError text

hansiu:
Hey,

Not much of an error, but could be helpful for a user to know that you meant 'zero' here :)

Have a nice day!


jdherman:
Thanks for catching this! I'll merge now (should not cause any conflicts with other open PRs).

Issue #213, opened 2018-10-19
Pandas friendly structure


ConnectedSystems:
Pandas DataFrames are the de facto standard for analysis purposes. It would be nice to have pandas friendly inter-operability out of the box.

willu47:
Hi @ConnectedSystems - thanks a lot for your contribution and welcome to SALib! Adding support for pandas "out of the box" seems like a nice idea, but I think we should extend your pull request to all the methods at once.  Also, I think an `results.as_df()` method could be a neat way, either to return a pandas DataFrame containing the sensitivity analysis results, or to provide a dict to be passed into a pandas DataFrame. The former option would require adding pandas to the list of SALib requirements, which I would prefer not to do, but if you have a compelling argument, I could be persuaded!

willu47:
Close by #214

PR #214, opened 2018-10-19
Function to create Pandas friendly dictionary

ConnectedSystems:
Working example to address the functionality I requested in Issue #213 for the Sobol method

Usage example:

```
import pandas as pd
from SALib.test_functions import Ishigami

X = saltelli.sample(problem, 1000)
Y = Ishigami.evaluate(X)
Si = sobol.analyze(problem, Y, print_to_console=True)

first, (idx, second) = sobol.Si_to_pandas_dict(Si, problem)
first_Si = pd.DataFrame(first, index=problem['names'])
second_Si = pd.DataFrame(second, index=idx)
```

### DIFF
--- a/SALib/analyze/delta.py
+++ b/SALib/analyze/delta.py
@@ -6,17 +6,17 @@ from scipy.stats import norm, gaussian_kde, rankdata
 import numpy as np
 
 from . import common_args
-from ..util import read_param_file
+from ..util import read_param_file, ResultDict
 
 
 def analyze(problem, X, Y, num_resamples=10,
             conf_level=0.95, print_to_console=False):
     """Perform Delta Moment-Independent Analysis on model outputs.
-    
+
     Returns a dictionary with keys 'delta', 'delta_conf', 'S1', and 'S1_conf',
     where each entry is a list of size D (the number of parameters) containing
     the indices in the same order as the parameter file.
-    
+
     Parameters
     ----------
     problem : dict
@@ -31,17 +31,17 @@ def analyze(problem, X, Y, num_resamples=10,
         The confidence interval level (default 0.95)
     print_to_console : bool
         Print results directly to console (default False)
-        
+
     References
     ----------
     .. [1] Borgonovo, E. (2007). "A new uncertainty importance measure."
            Reliability Engineering & System Safety, 92(6):771-784,
            doi:10.1016/j.ress.2006.04.015.
-           
+
     .. [2] Plischke, E., E. Borgonovo, and C. L. Smith (2013). "Global
            sensitivity measures from given data." European Journal of
            Operational Research, 226(3):536-550, doi:10.1016/j.ejor.2012.11.047.
-           
+
     Examples
     --------
     >>> X = latin.sample(problem, 1000)
@@ -61,7 +61,7 @@ def analyze(problem, X, Y, num_resamples=10,
     Ygrid = np.linspace(np.min(Y), np.max(Y), 100)
 
     keys = ('delta', 'delta_conf', 'S1', 'S1_conf')
-    S = dict((k, np.zeros(D)) for k in keys)
+    S = ResultDict((k, np.zeros(D)) for k in keys)
     if print_to_console:
         print("Parameter %s %s %s %s" % keys)
 
--- a/SALib/analyze/delta.py
+++ b/SALib/analyze/delta.py
@@ -62,6 +62,8 @@ def analyze(problem, X, Y, num_resamples=10,
 
     keys = ('delta', 'delta_conf', 'S1', 'S1_conf')
     S = ResultDict((k, np.zeros(D)) for k in keys)
+    S['names'] = problem['names']
+
     if print_to_console:
         print("Parameter %s %s %s %s" % keys)
 
--- a/SALib/analyze/dgsm.py
+++ b/SALib/analyze/dgsm.py
@@ -6,17 +6,17 @@ from scipy.stats import norm
 import numpy as np
 
 from . import common_args
-from ..util import read_param_file
+from ..util import read_param_file, ResultDict
 
 
 def analyze(problem, X, Y, num_resamples=1000,
             conf_level=0.95, print_to_console=False):
     """Calculates Derivative-based Global Sensitivity Measure on model outputs.
-    
+
     Returns a dictionary with keys 'vi', 'vi_std', 'dgsm', and 'dgsm_conf',
     where each entry is a list of size D (the number of parameters) containing
     the indices in the same order as the parameter file.
-          
+
     Parameters
     ----------
     problem : dict
@@ -32,7 +32,7 @@ def analyze(problem, X, Y, num_resamples=1000,
         The confidence interval level (default 0.95)
     print_to_console : bool
         Print results directly to console (default False)
-        
+
     References
     ----------
     .. [1] Sobol, I. M. and S. Kucherenko (2009). "Derivative based global
@@ -65,7 +65,7 @@ def analyze(problem, X, Y, num_resamples=1000,
 
     # First order (+conf.) and Total order (+conf.)
     keys = ('vi', 'vi_std', 'dgsm', 'dgsm_conf')
-    S = dict((k, np.zeros(D)) for k in keys)
+    S = ResultDict((k, np.zeros(D)) for k in keys)
     if print_to_console:
         print("Parameter %s %s %s %s" % keys)
 
--- a/SALib/analyze/dgsm.py
+++ b/SALib/analyze/dgsm.py
@@ -66,6 +66,8 @@ def analyze(problem, X, Y, num_resamples=1000,
     # First order (+conf.) and Total order (+conf.)
     keys = ('vi', 'vi_std', 'dgsm', 'dgsm_conf')
     S = ResultDict((k, np.zeros(D)) for k in keys)
+    S['names'] = problem['names']
+    
     if print_to_console:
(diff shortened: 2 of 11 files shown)


willu47:
Hi @ConnectedSystems - thanks a lot for your contribution and welcome to SALib! Adding support for pandas "out of the box" seems like a nice idea, but I think we should extend your pull request to all the methods at once. Also, I think an results.as_df() method could be a neat way, either to return a pandas DataFrame containing the sensitivity analysis results, or to provide a dict to be passed into a pandas DataFrame. The former option would require adding pandas to the list of SALib requirements, which I would prefer not to do, but if you have a compelling argument, I could be persuaded!

ConnectedSystems:
Thanks for the welcoming @willu47. 

Not sure how I feel about returning a DataFrame directly. Not only does it add a dependency (as you noted) but introduces tighter coupling with Pandas - not sure how I feel about this currently, but I'm erring on avoiding it. Currently I think just providing the data in a structure that is friendly to Pandas is okay - this way the user is not tied specifically to Pandas but can use any library that has adopted Pandas-like interfaces. That said, I can envision such methods being very convenient.

Is there any talk currently of implementing classes to address extensibility of the different methods and results? The current implementation doesn't use classes which makes implementing methods like `results.as_df()` (or `results.as_dict()`) a little less straight-forward than I'm used to (not saying it's not possible!). 

If there's a push towards re-organising things (or at least the result object) into classes I'm certainly keen to help where I can. Otherwise I can perhaps provide an example of how these methods might be implemented in without classes. 

Happy to keep discussing.

willu47:
I am keen to re-organise the library to reduce duplication of code and to ease maintainability. I can help coordinate this effort and provide feedback, with some effort towards coding/refactoring etc. I think @jdherman is keen to ensure that the library retains a simple structure, with separate modules for the methods across sample and analysis. I think that this _is_ compatible with some re-organisation effort - namely around the CLI (see #216), handling of 'problem' files, which define the number of parameters etc. For example, some methods support the grouping of parameters, some don't.

A starting point could be to write a class which can wrap around the sample and analyze methods (I  made a beginning using a strategy pattern within the Morris sample method which could be useful approach to separating the method implementations and the input/output admin) - and then refactor and tidy up once we are happy with the approach. This could then enable the `as_df`, `as_dict` method approaches etc.

jdherman:
Thanks @ConnectedSystems for the contribution, and @willu47 for the thoughtful response.

Big picture - a few years ago we discussed the possibility of moving to a fully object-oriented structure, but decided against it because it would be a major rewrite (including all tests, etc.) I know the current structure is not as clean as it could be - but it is easy to read the methods separately and understand how they work.

For this specific idea of converting the results to a pandas dataframe, I think that is a great idea and I don't see too much of a problem of adding a pandas dependency. 

In terms of where this goes in SALib, I'm not sure it should go inside the method-specific classes. Instead, what if there was a generic utility function (added to this file: https://github.com/SALib/SALib/blob/master/SALib/util/__init__.py) that accepted a dict of results and converted it to a dataframe? So the usage would be:

```python
from SALib.test_functions import Ishigami
from SALib.util import results_to_df

X = saltelli.sample(problem, 1000)
Y = Ishigami.evaluate(X)
Si = sobol.analyze(problem, Y, print_to_console=True)
Si_df = results_to_df(Si)
```

The challenge in writing the `results_to_df` function would be detecting the format of the results, possibly from the dict keys. The second-order results from Sobol may need to be a special case within this function.

IMO that would be the cleanest way to implement this without major structural changes, even though it is a little bit of a tack-on. What do you think?

ConnectedSystems:
Having a utility function is possibly the most pragmatic approach but at the same time, if I understand you (@jdherman) correctly, I can foresee that implementation being brittle - any changes to the analysis functions (or the addition of a new function) requires a change to the utility function and that might result in a bloated function signature/file. This deviates from the current situation where each analysis module is self-contained.

e.g.
```
# Inside util module
def results_to_df(results):
    # if results are from morris
    #     .... some conversion
    # elif results are from FAST:
    #     ... some conversion
    # ... and so on
```

Of course the conversion functions can be defined inside the separate `analysis` modules:

```
def results_to_df(results):
    # if results are from morris
    #     from SALib.analyze import morris
    #     return morris.to_df(results)
    # elif results are from FAST:
    # ... and so on
```

This approach still requires the separate utils file to be updated whenever changes happen.

Instead of the above, how about using a custom dict class? The relevant conversion function could then be attached as an object property when the result dictionary is created.

Example:

```
# from somewhere else - using the `utils` module for this example
# subclass is really only here to allow for custom method to be attached
# but could be expanded to be a full custom results class
class ResultDict(dict):
    # etc, etc
```

And within each module (`Sobol` in this example)

```
from SALib.util import ResultDict

def result_to_df(self):
    # Function to convert result dict into collection of Pandas DataFrame
    # ... see `Si_to_pandas_dict` in current PR for example that could be modified for this...
    # For consistency, could make all conversion functions return tuples of DataFrames
    return first, second, total


def create_Si_dict(D, calc_second_order):
    # Creates dictionary based on subclass above
    S = ResultDict((k, np.zeros(D)) for k in ('S1', 'S1_conf', 'ST', 'ST_conf'))

    # ... other stuff ...

    # attach conversion method
    S.to_df = result_to_df

    return S
```

Usage:
```
from SALib.test_functions import Ishigami

X = saltelli.sample(problem, 1000)
Y = Ishigami.evaluate(X)
Si = sobol.analyze(problem, Y, print_to_console=True)  # now returns ResultDict
first, second, total = Si.to_df()
```

This way conversion details and functions are contained within the module in question and conversions don't require a (possibly complicated) sequence of guesses as to what kind of results the dictionary holds.

willu47:
I think that's a great suggestion @ConnectedSystems. I'd be quite happy with a custom `ResultDict` class.  As you say, this would keep details of the implementation with the modules, but provide a consistent interface for the users.

ConnectedSystems:
Great! And thanks for the feedback. I will try to make time for this during the week.

jdherman:
Thanks @ConnectedSystems , this is a good suggestion. I am still wondering if this would be a lot of code to do something fairly simple. For all of the methods except Sobol, the results are just in the form `{'key1': array, 'key2': array}` where the arrays are values for each parameter. Even though the keys are different for each method, it should be easy to create a dataframe where the rows are the parameters (using string names) and the columns are those keys, if we just take them from dict.keys() without worrying about what they are.

The exception is the Sobol S2 and S2_conf, because they are DxD matrices rather than vectors. But given that this is the only special case, do you think a simple util function could still work here?

ConnectedSystems:
Because I subclass the base dict class, the ResultDict is - for all intents and purposes - a dictionary, so current behaviour should not break.

I've committed my changes - and I saw them appear - but now they've disappeared. Not sure what's happening.

jdherman:
Thanks! Looks simple and concise, and shouldn't change behavior.

This got caught up in the github outage yesterday (which might still be going on). Let's wait and see if the travis tests run, we might have to be patient for that.

ConnectedSystems:
Thanks for letting me know, @jdherman, I wasn't aware that github itself was having issues. Explains why you seemed to be especially concerned about maintainability ;)

I'd like to spend some more time cleaning up some inconsistencies between the DataFrames anyway - specifically including the parameter names as index keys.

ConnectedSystems:
I may have to finish this up over the weekend. 

I was trying to make everything consistent in that the DataFrames would have the parameter names as the index. I found in the process 1) that some methods include the parameter names in the dict, others don't, and 2) the Fractional Factorial method is another special case.

Issue 1 was easy to resolve, but 2 will take a little longer (I don't think too long, just longer than I have at the moment). I'd also like to add some tests so please hold off on the merge for now.

Enjoy the rest of the week

ConnectedSystems:
Hi @willu47 @jdherman 

Could I get some guidance on what would be preferred?

The Sobol and FF methods both return interaction sensitivities. Previously, the parameter names for Sobol second order sensitivities returned a tuple of the parameter pairs: e.g. (x1, x2)

However the FF method returned a string that was concatenated: e.g. "x1x2"

These get used as the DataFrame index:

```python
Si = analyze(problem, X, Y, second_order=True, print_to_console=False)

main_effect, inter_effect = Si.to_df()

inter_effect

# Outputs:
#         IE
# x1x2	0.000000
# x1x3	11.843525
# x2x3	0.000000
# x1dummy_0	0.000000
# x2dummy_0	11.843525
# x3dummy_0	0.000000
```

I feel tuples as indices are much clearer:

```python
# Outputs:
#         IE
# (x1, x2)	0.000000
# (x1, x3)	11.843525
# (x2, x3)	0.000000
# (x1, dummy_0)	0.000000
# (x2, dummy_0)	11.843525
# (x3, dummy_0)	0.000000
```

Current implementation uses a concatenated string for both FF and Sobol so that things are consistent. but I'd like to make them both use tuples instead though.

Thoughts?

jdherman:
Agreed, tuple would be better. I didn't realize FF was using concatenated strings. If this is changed for consistency, we may have to revise some of the FF unit tests too (not sure if they will break).

ConnectedSystems:
Done and done, ready for your review and approval.

A minor issue I came across is inconsistent or duplicate naming scheme for analysis and sample methods, e.g. 

```python
from SALib.sample import morris, ff, fast_sampler
from SALib.analyze import morris, ff, fast
```

But this is a separate issue that may get swept up in #216

willu47:
@ConnectedSystems - sorry for the delay in reviewing this. I should have time soon to sign this off!

PR #222, opened 2018-11-16
Consolidated CLI entry points

ConnectedSystems:
Added a single entry point for all SALib analysis and sampling methods.

Example usage:

```bash
salib sample fast_sampler -n 1000 -p ./SALib/test_functions/params/Ishigami.txt -o model_input.txt --delimiter=' '  --precision=8  -M 4 --seed=100
```

All sampling and analysis method files now include standard functions (`cli_parse` and `cli_action`) to support the consolidated entry point.

Tests to check that the CLI works as expected have also been added.

Example scripts have also been updated. The catch is that they are currently commented out. 
Tested CLI entry point installation with pyscaffold v2.5 and it works as expected (under Windows).


ConnectedSystems:
So that took a lot longer than I thought it would. Took me a while to get tests to run without issue across Windows, Linux, and Travis on both Python 2 and 3.

ConnectedSystems:
@willu47 @jdherman 

I believe this is ready for your review.

Issue #152, opened 2017-06-29
delta.analyze fails with LinAlgError for small sample sizes


willu47:
See #151 for discussion on this error.

Use the test file below to reproduce error:

```python
from SALib.analyze.delta import analyze
import numpy as np

def test_bug_delta_linalg():

    problem = problem = {
                'num_vars': 3,
                'names': ['K_IO_AOBden', 'K_SO_AOBden', 'K_SNO_AOBden'],
                'bounds': [[0.01, 11], [0.01, 13.2], [0.08, 4.3]]
                         }

    X = np.array([[2.12591411, 1.8200793, 0.10082878],
                  [6.21754537, 5.85214066, 2.8846046],
                  [8.01209459, 12.95017617, 1.00034018],
                  [10.72587215, 9.03629573, 1.88932555],
                  [4.17005179, 1.30648717, 3.7287254],
                  [3.24299291, 3.44273292, 4.01808892],
                  [8.80676551, 10.03481953, 1.43068963],
                  [0.29322542, 6.80514303, 2.43057428]])

    Y = np.arange(8)

    analyze(problem, X, Y, num_resamples=100)
```

ConnectedSystems:
Based on discussion here, suggest testing for singular matrix and throwing error indicating an inadequate sample size:

```python
from __future__ import division

from SALib.analyze.delta import analyze
import numpy as np
import sys


def test_bug_delta_linalg():

    problem = problem = {
                'num_vars': 3,
                'names': ['K_IO_AOBden', 'K_SO_AOBden', 'K_SNO_AOBden'],
                'bounds': [[0.01, 11], [0.01, 13.2], [0.08, 4.3]]
                         }

    X = np.array([[2.12591411, .8200793, 0.10082878],
                  [6.21754537, 5.85214066, 2.8846046],
                  [8.01209459, 12.95017617, 1.00034018],
                  [10.72587215, 9.03629573, 1.88932555],
                  [4.17005179, 1.30648717, 3.7287254],
                  [3.24299291, 3.44273292, 4.01808892],
                  [8.80676551, 10.03481953, 1.43068963],
                  [0.29322542, 6.80514303, 2.43057428]])
    
    # This would go into `analyze()`
    if np.linalg.cond(X) < 1/np.finfo(X.dtype).eps:
        raise ValueError("Singular matrix detected - sample size may be too small")
    # End if

    Y = np.arange(8)
    
    analyze(problem, X, Y, num_resamples=100)
```

[...]

PR #225, opened 2019-03-04
Raise error if singular matrix is detected

ConnectedSystems:
Raise error with message indicating sample size may be too small

Address PR #152

### DIFF
--- a/SALib/analyze/delta.py
+++ b/SALib/analyze/delta.py
@@ -48,11 +48,6 @@ def analyze(problem, X, Y, num_resamples=10,
     >>> Y = Ishigami.evaluate(X)
     >>> Si = delta.analyze(problem, X, Y, print_to_console=True)
     """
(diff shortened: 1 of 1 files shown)


willu47:
@ConnectedSystems - I note the current tests fail as this new error is raised for the sample sizes of 8,000 and 10,000!

ConnectedSystems:
Hi Will,

Yes - was looking into it but then got distracted with the other request - will respond soon

antonia-had:
Hi, was wondering what the status of this is? I keep getting it for large samples (10000) also, I believe it might also be caused if the model output is not sufficiently variant/contains many zeros (the case with mine)

jdherman:
Hi @antonia-had , I just pushed a new release (v1.3.4) to pypi. This includes the bug fixes that were merged in this PR.

PR #235, opened 2019-04-23
Updated to work with NAN

j-chacon:
Delta, rbd_fast and sobol analysis routines were modified to admit NAN values in the model results.

### DIFF
--- a/src/SALib/analyze/delta.py
+++ b/src/SALib/analyze/delta.py
@@ -60,7 +60,7 @@ def analyze(problem, X, Y, num_resamples=10,
     # equal frequency partition
     M = min(np.ceil(N ** (2 / (7 + np.tanh((1500 - N) / 500)))), 48)
     m = np.linspace(0, N, M + 1)
-    Ygrid = np.linspace(np.min(Y), np.max(Y), 100)
+    Ygrid = np.linspace(np.nanmin(Y), np.nanmax(Y), 100)
 
     keys = ('delta', 'delta_conf', 'S1', 'S1_conf')
     S = ResultDict((k, np.zeros(D)) for k in keys)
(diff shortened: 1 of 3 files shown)


willu47:
Hi @j-chacon - many thanks for the pull request. Please could you create an issue which describes the issue this pull request addresses (see our [contributing](https://github.com/SALib/SALib/blob/master/CONTRIBUTING.md) guidelines) as that makes it easier to review your contribution.

I also note that the tests are failing due to a Travis update.  I'll look into fixing those now.

j-chacon:
Hi @willu47 

I have opened the issue #237. I hope it's clear now.

willu47:
Hi @j-chacon - please could you add some tests which demonstrate expected behaviour when the methods are passed results data containing `NaN` values?

j-chacon:
HI @willu47, the pull request has been updated to raise a ValueError if NAN values are found in the model results, for all methods. In the case of `sobol.analyze` a flag (`ignore_nans`) that allows calculating the indexes with NAN values was introduced. The flag is by default set to `False`.

In addition, tests were added in `test_functions_with_NAN.py` to demonstrate the functionalities of the new features.

joaoantoniocardoso:
This would be very helpful for my work, it's tough to keep some physical models stable when varying parameters, often ending up with NaNs from model simulation, falling today's SAlib implementation.

tupui:
I don't think we should do this besides raising in case of NaN values (see the linked issues). The guidelines often tell users to use samples like Sobol' or Saltelli's sampling with specific matrices. Removing points create sampling imbalance. It should be up to the user to do something with their sampling knowingly. There is just to many different cases to cover and I don't think we could easily come up with an API for it or have some detection of some sort to do something sensible.

PR #243, opened 2019-05-12
Updated import of scipy comb function

ConnectedSystems:
Importing from scipy.misc now deprecated.

Importing from scipy.special instead


jdherman:
Thanks!

PR #246, opened 2019-05-20
Improved efficiency of summing distances in local optimization

JoerivanEngelen:
I did some profiling on a slightly modified morris sample example (master/examples/morris/morris.py). 
Only change I made to the example:

```
param_values = sample(problem, N=80, grid_jump=2, num_levels=4,
                      optimal_trajectories=79)
```

The main bottleneck is in `sum_distances` within `LocalOptimisation`. My laptop originally spent 85 seconds in this function and with this improvement it spent 44 seconds, a near 50% speed gain! 

Main culprits were the unnecessary list comprehensions, even though a call to zip suffices here.
Code is also slightly cleaner now, in my opinion :-)

Profiling was done in Spyder, I added the profiling files if you are interested.

### DIFF
--- a/src/SALib/sample/morris/local.py
+++ b/src/SALib/sample/morris/local.py
@@ -108,11 +108,9 @@ class LocalOptimisation(Strategy):
         are only used in a relative way. Purely summing distances would lead
         to the same result, at a perhaps quicker rate.
         """
-        combs_tup = np.array(tuple(combinations(indices, 2)))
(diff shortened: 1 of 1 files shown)


ConnectedSystems:
Hi @JoerivanEngelen would you mind sharing the files you used to do the profiling?

I'll take this opportunity to see if I can make some changes as well.

Issue #131, opened 2017-01-20
Add new method - New Sampling Strategy for Method of Morris


willu47:
[Khare et al. (2015)](http://www.sciencedirect.com/science/article/pii/S1364815214003399) describe an improved method for sampling for the method of morris.

JoerivanEngelen:
I'm currently looking into this, implementing it should not be too hard. 
(Was not really happy with the uniformity of my optimized samples)

If I understand it correctly, the main improvement is that you ensure that the first points of the trajectories are uniformly distributed across parameters. 
(Isn't that the same as Latin Hypercube sampling?) 

The rest of the method in the paper is what we call "brute force optimization" in SALib. I expect to work on this tomorrow. 

FYI: Original (MATLAB) code can be found here:
https://abe.ufl.edu/faculty/carpena/software/SUMorris.shtml

willu47:
Okay, do note the license terms:

> These Matlab (R) packages were developed by Drs. Yogesh Khare and Rafael Muñoz-Carpena. This program is distributed as Freeware/Public Domain under the terms of GNU-License. If the program is found useful, the authors ask that acknowledgment is given to its use in any resulting publication and the authors notified. The source code is available from the authors upon request.

PR #247, opened 2019-05-23
First attempt at incorporating sampling for uniformity (Morris)

JoerivanEngelen:
For method, see Khare et al. 2015 _A multi-criteria trajectory-based parameter sampling strategy for the screening method of elementary effects_

I thought I'll leave it as a pull request here, so we can discuss the code.

I had to hack the sampling strategy quite a bit to get this one to work. Goal was to use as much of the original code and functions as possible. Currently the variable names are quite bad and the code can be quite confusing. I also had to override some of the checks.
Most confusing is that the RepeatedSampling class inherits the Strategy class with:
sampling4uniformity=num_pars 
and 
num_pars=k_choice

For this to be more understandable, probably the original code has to be refactored a bit. 

You can try it out yourself by for example:
```
param_values = sample(problem, N=16, grid_jump=2, num_levels=4,
                      optimal_trajectories=None, 
                      sample4uniformity = 300)
```

To do:

- [ ]  e-mail original authors we are working on this 
- [ ]  add better docstrings
- [ ]  better variable names
- [ ]  add checks, raise ValueErrors/Warnings
- [ ]  test whether this implementation actually provides a more uniform parameter spread
- [ ]  support groups
- [ ]  refactor?

### DIFF
--- a/src/SALib/sample/morris/__init__.py
+++ b/src/SALib/sample/morris/__init__.py
@@ -34,6 +34,7 @@ import numpy.random as rd
 from . gurobi import GlobalOptimisation
 from . local import LocalOptimisation
 from . brute import BruteForce
+from . repeatedsampling import RepeatedSampling
 
 from . strategy import SampleMorris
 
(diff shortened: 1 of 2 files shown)


JoerivanEngelen:
This should fix #131

JoerivanEngelen:
I see the most of the tests now fail because no grid_jump is provided in the tests, which raises the question: how did the tests for previous commits not fail?

willu47:
> I see that most of the tests now fail because no grid_jump is provided in the tests, which raises the question: why didn't they fail in previous commits?

The grid jump argument was removed in an earlier release, as it is was computed directly from the `num_levels` parameter.  This is why it doesn't appear in the tests.

Issue #241, opened 2019-05-09
Singular matrix issue when computing delta indeces


ivan-marroquin:
Hi there,

I noted that there are two closed incidents on this topic (#152 and #1510), as well a recent one (#240). But I believe that the issue I'm having is different to those that were reported.

I am using Python 3.6.5 on a windows machine and SALib 1.3.3

I conducted a series of tests using four input parameters to derive a single output. There are four scenarios, so I want to evaluate Delta indices in each scenario for different bootstrap levels by using the following code (note that I consider "resamples" option in Delta index as the number of samples to be taken in bootstrap procedure):

++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
problem= {
        'num_vars': 5,
        'names': ['input_1', 'input_2', 'input_3', 'input_4', 'input_5'],
        'bounds': [[min_input_1, max_input_1],
                   [min_input_2, max_input_2],
                   [min_input_3, max_input_3],
                   [min_input_4, max_input_4],
                   [min_input_5, max_input_5]],
        }

Si= []

filename= 'delta_moment_3d_orig_n_noise_class_'
figure_name= 'delta_moment_3d_orig_n_noise_class_'

label= 'Free noise wedge model'

attribute_names= ['input_1', 'input_2', 'input_3', 'input_4', 'input_5']

bootstrap= 0    

print("Perform Delta-Moment Independent Test\n")

for count in range(0,len(classification_names)): # Test different scenarios
    filename= filename + np.str(count + 1) + '.csv'
    figure_name= figure_name + np.str(count + 1)
    
    output_file= open(path + filename, 'w')
    
    for iters in range(0,4): # Test different bootstrap levels
        bootstrap += 200
    
        print("using a bootstrap sampling of {} at iteration {} and count {}\n".format(bootstrap, iters, count))

        Si= delta.analyze(problem, attribute_data, classification_data[:,count], num_resamples= bootstrap, print_to_console= True)

        if (bootstrap == 200):
            field_names= []
        
            for key in Si:
                field_names.append(str(key))
            
            output_file.write("%s %s %s %s %s %s %s %s %s %s %s\n" %("attribute name", ',', field_names[0], ',',
                                                                     field_names[1], ',', field_names[2], ',', field_names[3], ',',
                                                                     "bootstrap level"))

        delta_r= Si.get('delta')
        delta_conf_r= Si.get('delta_conf')
        sobol_r= Si.get('S1')
[...]

Issue #240, opened 2019-05-08
Singular matrix when calculating Delta indices


antonia-had:
I'm submitting this as a new issue as #225 is closed.

So I updated to 1.3.4 as per @jdherman's suggestion and I still get the same singular matrix error (or, rather, the updated message). It’s doing it for both 1000 samples and 10000 samples of the same analysis done in different ways. There’s a lot of zeros in my output, so I thought that’s what’s causing it. 
I went to manually replicate the calculation where it breaks down. The script loops through every parameter and estimates the difference with the density function in the _calc_delta_ function below (where Y is the entire output and X is the samples for one of the parameters). The error is always produced when estimating the density function of Y at the specific X values (fyc = …)
In my case the Y was water shortage, and one parameter specifically is an on/off switch that [what it does doesn’t matter] but results in 0 shortages. gaussian_kde won’t work for an array of zeros (giving a singular matrix error) but the parameter indeed shifts the density function rather significantly and the difference should be added to d_hat. I get an error where I should actually be getting a very high index for that parameter. Does that make sense?
```
def calc_delta(Y, Ygrid, X, m):
    N = len(Y)
    fy = gaussian_kde(Y, bw_method='silverman')(Ygrid)
    xr = rankdata(X, method='ordinal')

    d_hat = 0
    for j in range(len(m) - 1):
        ix = np.where((xr > m[j]) & (xr <= m[j + 1]))[0]
        nm = len(ix)
        fyc = gaussian_kde(Y[ix], bw_method='silverman')(Ygrid)
        d_hat += (nm / (2 * N)) * np.trapz(np.abs(fy - fyc), Ygrid)

    return d_hat
```
Lastly, I don't know if it would be valuable to still report on the delta values of the other parameters even if one of them causes an error (for whatever reason). In my case all other parameters work fine, but this pesky one halts the whole thing.

antonia-had:
Update:
This fixes the issue (checks if any value in Y[ix] is non-zero before estimating density function)
```
        if Y[ix].any():
            fyc = gaussian_kde(Y[ix], bw_method='silverman')(Ygrid)
            d_hat += (nm / (2 * N)) * np.trapz(np.abs(fy - fyc), Ygrid)
        else:
            d_hat += (nm / (2 * N)) * np.trapz(np.abs(fy), Ygrid)
```

ConnectedSystems:
Hi @antonia-had 

Would you be able to share an example dataset? I'd like to add this situation as a test case.

Thanks

[...]

PR #251, opened 2019-06-06
Handle singular value matrix cases

ConnectedSystems:
Handle cases where subset of ranked data results in a matrix with identical values leading to a singular value matrix exception when using the Delta analysis method.

Based on the proposed fix by @antonia-had.

Includes additional regression test for the Delta method.
This test could be simplified further to recreate past fail condition (currently relies on a fairly large array to be created).

Suggest @jdherman review the new general exception message advising users to raise an issue with the SALib team

Also suggest discussion around whether this PR adequately resolves the situations as encountered in #240 and #241 as I am unable to comment on the possible consequences of using the absolute values of a gaussian kernel-density estimate in these specific cases, and possible side-effects generally.

(pinging @antonia-had and@ivan-marroquin so they are aware this PR has been submitted)


jdherman:
Looks good to me, thanks @ConnectedSystems and @antonia-had. I will merge now and upload the new version to pypi.

jdherman:
Looks good to me, thanks @ConnectedSystems and @antonia-had. I will merge now and upload the new version to pypi.
(sorry for the confusion I closed it by accident the first time)

PR #264, opened 2019-09-19
Further performance improvements

ConnectedSystems:
Additional changes for performance for delta and DGSM methods. 

Reduction in run time by up to 30%

### DIFF
--- a/src/SALib/analyze/delta.py
+++ b/src/SALib/analyze/delta.py
@@ -58,12 +58,12 @@ def analyze(problem, X, Y, num_resamples=10,
         raise RuntimeError("Confidence level must be between 0-1.")
 
     # equal frequency partition
-    M = min(np.ceil(N ** (2 / (7 + np.tanh((1500 - N) / 500)))), 48)
+    M = min(int(np.ceil(N ** (2 / (7 + np.tanh((1500 - N) / 500))))), 48)
     m = np.linspace(0, N, M + 1)
     Ygrid = np.linspace(np.min(Y), np.max(Y), 100)
 
     keys = ('delta', 'delta_conf', 'S1', 'S1_conf')
-    S = ResultDict((k, np.zeros(D)) for k in keys)
+    S = ResultDict((k, np.empty(D)) for k in keys)
     S['names'] = problem['names']
 
     if print_to_console:
@@ -119,7 +119,7 @@ def calc_delta(Y, Ygrid, X, m):
 
 
 def bias_reduced_delta(Y, Ygrid, X, m, num_resamples, conf_level):
-    d = np.zeros(num_resamples)
+    d = np.empty(num_resamples)
     d_hat = calc_delta(Y, Ygrid, X, m)
 
     N = len(Y)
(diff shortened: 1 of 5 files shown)


src/SALib/analyze/delta.py
```diff
@@ -58,26 +58,27 @@ def analyze(problem, X, Y, num_resamples=10,
...
 
     # equal frequency partition
-    M = min(np.ceil(N ** (2 / (7 + np.tanh((1500 - N) / 500)))), 48)
+    M = min(int(np.ceil(N ** (2 / (7 + np.tanh((1500 - N) / 500))))), 48)
     m = np.linspace(0, N, M + 1)
     Ygrid = np.linspace(np.min(Y), np.max(Y), 100)
 
     keys = ('delta', 'delta_conf', 'S1', 'S1_conf')
-    S = ResultDict((k, np.zeros(D)) for k in keys)
+    S = ResultDict((k, np.empty(D)) for k in keys)
```
willu47: Do note that using `np.empty` can cause issues if the array is not subsequently filled with values. This caused some hard to find bugs in the past.

src/SALib/analyze/delta.py
```diff
@@ -134,20 +136,22 @@ def sobol_first(Y, X, m):
     xr = rankdata(X, method='ordinal')
     Vi = 0
     N = len(Y)
+    Y_mean = Y.mean()
     for j in range(len(m) - 1):
         ix = np.where((xr > m[j]) & (xr <= m[j + 1]))[0]
         nm = len(ix)
-        Vi += (nm / N) * (Y[ix].mean() - Y.mean()) ** 2
+        Vi += (nm / N) * (Y[ix].mean() - Y_mean)**2
```
willu47: Spaces around operators (PEP8)

ConnectedSystems: Sorry Will, I disagree with this interpretation.

PEP8 does not enforce spacing around the power operator (see [here](https://wiki.fysik.dtu.dk/ase/development/python_codingstandard.html#run-autopep8-py-on-your-code) for one example)

You wouldn't write `Y = x^2` as `Y = x 2` 
(i.e. you wouldn't put a space between the coefficient and its power)

src/SALib/analyze/delta.py
ConnectedSystems: Yes, I am aware of this.

That said, the implementations here should all subsequently fill in the arrays and use of `empty` has a slight performance benefit.

If `D` values were not filled, I think that's sign of a larger issue.
As it stands though I'm happy to revert everything back to using `np.zeros` although it will pain me to do so...

src/SALib/analyze/delta.py
ConnectedSystems: How about this as a compromise?

`Vi += (nm / N) * ((Y[ix].mean() - Y_mean)**2)`

Suggested in:

https://github.com/PyCQA/pycodestyle/issues/292#issuecomment-44591870

willu47: Your compromise looks good. Thanks for looking into this!

src/SALib/analyze/delta.py
willu47: Okay.

src/SALib/analyze/delta.py
willu47: Your compromise looks good. Thanks for looking into this!

ConnectedSystems:
Applied requested changes, principally reverting use of `np.empty` to `np.zeros`

PR #265, opened 2019-09-19
Radial oat method

### LINKED ISSUE #132: Add new Method - Radial sampling strategy for Method of Morris
[Paper describing the radial sampling strategy](http://www.sciencedirect.com/science/article/pii/S0010465510005321)

ConnectedSystems:
Address #132 

Radial OAT sampling and analysis method.
Includes Elementary Effects and Jansen's sensitivity estimation.

Still a work in progress and not ready for merge.

Requires:
* Additional testing and code review (by someone other than me)
* Clean up/additional docstrings
* CLI interface functions
* Tests for CI
* Example scripts


src/SALib/sample/radial.py
```diff
@@ -0,0 +1,83 @@
...
+
+    grp_start = 0
+    for i in range(perturbations.shape[0]):
+        mod = np.diag(perturbations[i])
+
+        np.copyto(sample_set[grp_start+1:grp_start+p+1], mod, where=mod != 0.0)
+        grp_start += p+1
+    # End for
+
+    # The above will look like:
```
willu47: Can you add this under an `Example` heading in the docstring?

src/SALib/sample/radial.py
```diff
@@ -0,0 +1,83 @@
...
+    discard = 4
+    R = N + discard
+
+    # Generate the 'nominal' parameter positions
+    # Total number of parameter sets = N*(p+1)
+    base_sequence = sobol_sequence.sample(R, p)
+    subsetted_base = base_sequence[discard:]
+    scale_samples(subsetted_base, bounds)
+
+    group = p+1
```
willu47: Add spaces around operators (PEP8)

src/SALib/analyze/radial_st.py
```diff
@@ -0,0 +1,119 @@
...
+
+__all__ = ['analyze']
+
+def analyze(problem: Dict, Y: np.array, sample_sets: int,
+            num_resamples: int = 1000,
+            conf_level: float = 0.95,
+            seed: Optional[int] = None) -> Dict:
+    """Global Sensitivity Index for `radial` approach.
+
+    Sample size must be of a compatible number, or else results
```
willu47: What is a compatible number? Can you provide guidance in the docstring here?

willu47 (changes requested):
Hi @ConnectedSystems. Many thanks for this pull request which tackles issue #132 

I've made some comments on syntax inline.

Is it worth adding this as a strategy within the Morris submodule?  And do the analyze methods differ significantly from the existing implementations?

As you state, there's still a fair amount of work on implementing tests for these new modules. It may be worth parameterising existing tests for the morris method with the different methods. Ideally, they should all produce the same values.

Happy to iterate on this.

ConnectedSystems:
> I've made some comments on syntax inline.

Thanks - for some I'm obviously reliant on a linter which usually handles styling for me - it appears not to be working for me at the moment but have addressed these manually.

> Is it worth adding this as a strategy within the Morris submodule?

I was going to ask about this. Wasn't sure how best cleanly incorporate it at first glance but will take a more in-depth look again tomorrow.

> And do the analyze methods differ significantly from the existing implementations?

I think they differ enough to warrant separate implementations but happy to be corrected on this. I plan to have a closer examination when I have more time later on.

> As you state, there's still a fair amount of work on implementing tests for these new modules. It may be worth parameterising existing tests for the morris method with the different methods. Ideally, they should all produce the same values.

The same values, as in, the results between Morris and Radial OAT should be identical?
From my initial testing, this is not the case but the rankings were the same.

src/SALib/analyze/radial_st.py
ConnectedSystems: Modified docstring to be clearer, with references

ConnectedSystems:
Still todo:

* Tests for CI
* CLI interface functions
* Example scripts

ConnectedSystems:
Excuse the delay on this - been busy with a truckload of other things.

Latest round of commits include an initial support for the CLI and changes to the docstrings to better address last review comments.

Preliminary tests produces results in line with what was reported in Campolongo et al. (2011), although I don't get the exact quoted index values - probably because I used a lower sample size.

Specifically, using the modified G-function (with 20 parameters) with the parameter values specified in Table 3 of the cited article above:

* Identified parameters of importance for `G_4`: 2, 8, 11, 18

G4_ee_results
G4_st_results

* Identified parameters for `G_10`: 2, 7, 8, 9, 10, 11, 12, 15, 18, 20

G10_ee_results

G10_st_results

ConnectedSystems:
Almost there - biggest job now is to add support for usage via CLI

ConnectedSystems:
Finalized tests and initial CLI support.

Still to do:

* Finalize example scripts for documentation, which will lead to final CLI support
* Clean up code where needed

ConnectedSystems:
This PR ready for review

willu47 (approved):

lbteixeira:
Hi, @ConnectedSystems.

In #320 you mentioned this topic, and I could help if you need. Do you have a more or less working code? Or perhaps I could start from scratch?

ConnectedSystems:
Hi @lbteixeira 

This specific implementation is working. If you care to take a look you can check out the branch in my repository. 
The branch name is `radial-oat-method` (https://github.com/ConnectedSystems/SALib/tree/radial-oat-method).

What would be nice is to make this compatible with the [Morris strategy class](https://github.com/SALib/SALib/blob/649ffbd12acd4c879afb1726f005fb8a4b5ce7b1/src/SALib/sample/morris/strategy.py#L56)

I have not looked into this in detail so I am not sure if the data structures are compatible and if not, what would be required to make them compatible.

While it would be nice to merge things, I also don't want you to spend too much time on this.

ConnectedSystems:
Closing in favour of #689

PR #267, opened 2019-10-01
Bump pyscaffold version

ecederstrand:
The current pyscaffold contains a vendored version of `setuptools_scm` that contains a bug in `list_files_in_archive` when the `setuptools` 41.2.0 `walk_revctrl()` function is called with default args. This breaks pip installation of random other packages in your requirements.txt with the following output because setuptools_scm registers itself as a plugin to setuptools:

```
writing pip-egg-info/python_bioformats.egg-info/PKG-INFO
writing requirements to pip-egg-info/python_bioformats.egg-info/requires.txt
writing top-level names to pip-egg-info/python_bioformats.egg-info/top_level.txt
writing dependency_links to pip-egg-info/python_bioformats.egg-info/dependency_links.txt
*** FileNotFoundError: [Errno 2] No such file or directory: ''
```

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -5,11 +5,11 @@
 SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
 PAPER         =
-BUILDDIR      = _build
+BUILDDIR      = ../build/sphinx/
 AUTODOCDIR    = api
 AUTODOCBUILD  = sphinx-apidoc
 PROJECT       = SALib
-MODULEDIR     = ../src/SALib
+MODULEDIR     = ../src/salib
 
 # User-friendly check for sphinx-build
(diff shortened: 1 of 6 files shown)


willu47:
Many thanks for this contribution! One quick question - did you check to see if there have been any other updates to the configuration files since previous versions, which we may need to implement in our `setup.cfg` and `setup.py` files?

ecederstrand:
No, I didn't check. How can I do that? I haven't actually used pyscaffold before. I just ended up here after a lengthy debug session :-)

willu47:
Oof, sorry to hear that. I'll have a quick look.

willu47:
I notice in the [changelog](https://pyscaffold.org/en/latest/changelog.html) of the pyscaffold project a number of important updates in version 3.1 and 3.2.  We can automate this using the `--update` flag of the pyscaffold `putup` command.

PR #269, opened 2019-10-09
Changed Sobol G-function to modified G-function

QianWanghhu:
Shift and curvature parameters are added.
Taken from Saltelli et al. (2010). 
Reference in docstring.
Changed docstring.

### DIFF
--- a/src/SALib/test_functions/Sobol_G.py
+++ b/src/SALib/test_functions/Sobol_G.py
@@ -23,24 +23,24 @@ def evaluate(values, a=None, delta=None, alpha=None):
 
     Parameters
     ----------
-    values : ndarray, input variables
-    a : np.array,
-    delta : np.array, of delta coefficients.
-    alpha : np.array, of alpha coefficients
+    values : ndarray
+        input variables
+    a : np.array
+        parameter values
+    delta : np.array
+        shift parameters
+    alpha : np.array
+        curvature parameters
 
     Returns
     -------
-    y : Scalar of G-function result.
+    Y : Result of G-function
     """
     if type(values) != np.ndarray:
         raise TypeError("The argument `values` must be a numpy ndarray")
-    if delta and not isinstance(delta, np.ndarray):
-        raise TypeError("The argument `delta` must be given as a numpy ndarray")
-    if alpha and not isinstance(alpha, np.ndarray):
-        raise TypeError("The argument `alpha` must be given as a numpy ndarray")
-
+  
     if a is None:
-        a = [0, 1, 4.5, 9, 99, 99, 99, 99]
+        a = np.array([0, 1, 4.5, 9, 99, 99, 99, 99])
 
     if delta is None:
         delta = np.zeros_like(a)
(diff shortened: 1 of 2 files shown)


QianWanghhu:
Still need to write tests.

QianWanghhu:
It's ready to be merged.

src/SALib/test_functions/Sobol_G.py
```diff
@@ -10,11 +10,57 @@
...
+
+    .. [1] Saltelli, A., Annoni, P., Azzini, I., Campolongo, F., Ratto, M., 
+           Tarantola, S., 2010. Variance based sensitivity analysis of model 
+           output. Design and estimator for the total sensitivity index. 
+           Computer Physics Communications 181, 259–270. 
+           https://doi.org/10.1016/j.cpc.2009.09.018
+
+    Parameters
+    ----------
+    values : ndarray
```
willu47: type here and below should be 
`values : numpy.ndarray`

src/SALib/test_functions/Sobol_G.py
```diff
@@ -26,40 +72,48 @@ def evaluate(values, a=None):
...
+        temp_y = (np.abs(2 * mod_x - 1)**alpha)
+        y_elements = ((1 + alpha) * temp_y + a) / (1 + a)
+        Y[i] = np.prod(y_elements)
 
     return Y
 
 
-def partial_first_order_variance(a=None):
+
+def partial_first_order_variance(a=None, alpha=None):
```
willu47: If this is a private function e.g. not intended to be called outside of the module, could you prepend `_` e.g. `def _partial_first_order_variance()`

tests/test_test_functions.py
```diff
@@ -97,3 +97,50 @@ def test_total_sensitivity_index():
                          0.222116249, 0.030859879, 0.032170899])
 
     assert_allclose(actual, expected, atol=1e-2, rtol=1e-6)
+
+
+def test_Modified_Sobol_G():
```
willu47: remove CAPITALS from function name

tests/test_test_functions.py
```diff
@@ -97,3 +97,50 @@ def test_total_sensitivity_index():
...
+    delta_values = [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8]
+    with raises(TypeError):
+        evaluate(parameter_values, delta=delta_values)
+
+    alpha_values = [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8]
+    with raises(TypeError):
+        evaluate(parameter_values, alpha=alpha_values)
+
+
+def test_Modified_Sobol_G_error_if_value_beyond_range():
```
willu47: It's worth testing the error messages too. Something like:
```python
with raises(ValueError) as ex:
    <function call which raises the error>
assert "The error message I expect" in str(ex)
```

ConnectedSystems: Hi @QianWanghhu and @willu47 

Note that with later versions of pytest the `in str(ex)` approach will error. The alternative is either:

```python
with raises(ValueError) as ex:
    <function call which raises the error>

assert "The error message I expect" in str(ex.value)

# OR

# `match` accepts regex if desired
with pytest.raises(ValueError, match="<expected error message>"):
    <function call which raises the error>
```

I had to adjust this a few months ago, see:
https://github.com/SALib/SALib/commit/9fe3c129ae5476c38b5c2d7622565635dabc287f

QianWanghhu: Hi @willu47 and @ConnectedSystems 
Thanks for your comments and suggestions which are very helpful. I will fix it accordingly.

QianWanghhu:
Hi @willu47, I have addressed the comments. The capitals in "Sobol G-function" are not changed in order to be consistent within the package.

Issue #234, opened 2019-04-21
OO Problem class & method compatibility checking


ConnectedSystems:
This will be a long read but I hope to spark some discussion around improvements to SALib in terms of usability in the longer term.

Before I start I'd like to thank everyone involved in making this package, it certainly has made doing this kind of work more accessible.

In using this package there are some pain points I have encountered which I'd like to show examples of, and suggest possible solutions. Some of the issues I raise below are relatively minor, but addressing these could improve usability.

Starting with the most minor:

1. Redundant attribute in "problem" specification

```python
problem = {
  'num_vars': 3,
  'names': ['x1', 'x2', 'x3'],
  'bounds': [[-np.pi, np.pi]]*3
}
```

I've always thought "why do I have to specify the number of parameters? It could be inferred from the number of names."

But I've noticed that some methods directly modify this value (in ff.py for example). This leads into the next suggestion of reconceptualizing how SALib represents the different sampling/analysis methods. Currently they are a set of functions which were developed by different contributors - and nothing wrong with that! But it does mean that methods are not always compatible (more on this later).

2. Objects for SALib

Previously it was suggested that a move towards a more object-oriented approach be made [as in here](https://github.com/SALib/SALib/issues/216#issuecomment-435647632). I previously pushed for keeping things as they currently are for the short-term but given the context here is for the medium/long term changes it's worth revisiting.

To implement my first suggestion would 1) require all methods to have the check implemented and create the `num_vars` element if it doesn't exist, 2) be reorganized into a class hierarchy from which the check is inherited, or 3) use a common decorator function that implements the check.

The first would introduce unnecessary duplication and complexity so that's off the table. The second is viable but would mean a significant overhaul of how things are implemented, with implications for the testing setup that is already in place. The third approach is what I'm currently leaning towards as it retains most of the current implementation (see further below for an example).

3. Mixing and matching methods

Currently users are able to mix and match different sampling and analysis methods without any indication that it is sane to do so. This leads to some confusion [as in here](https://github.com/SALib/SALib/issues/218) and [here](https://github.com/SALib/SALib/issues/224); addressing this pain point would go a long way to making SALib more user-friendly.

4. An example

To bring all the points above together, I'll adapt the canonical example from the readme.

I should also ask at this point if there's been any previous discussion around what the envisioned OO approach looks like.

The below is only an idea, none of the below has been tested, implemented, or checked for sanity, but I think it is doable.

```python
# This, with or without `num_vars`, would be a valid spec
problem = {
[...]

Issue #193, opened 2018-02-07
Generating correlated distributions of data?


spizwhiz:
Hello,

I have sampled  data with the saltelli sampler, but so far have only been able to generate several univariate distrubtions.

I am wondering if it is possible to generate a multivariate distrubtion with a correlation matrix?

Thank You!

jdherman:
Hi @spizwhiz ,
All of the methods in SALib currently assume independent parameters.

You could write your own modification to impose a correlation on a sample, but I'm not sure if the rest of the methods would still work as intended.

There are many papers that discuss this issue. For example:
https://www.sciencedirect.com/science/article/pii/S0951832005002231
To get these methods into SALib, we would need a pull request from an expert!

spizwhiz:
Thank you.  

I am afraid I am not that expert.

Looking at the paper, it seems like their approach was just to group correlated parameters into a sort of lumped parameter, and analyzed the sensitivity of the model to the independent groups of parameters.  This makes sense for their application, but I am not sure  it will apply to my model (richards equation based), as I am testing sensitivity to 4-6 parameters, with correlation coefficients ranging from +\- .1 to +\- .9, most of them absolute value greater than .5.

I'd be grateful if you had suggestions (or even relevant citations) for using SALib or other methods for sensitivity analysis on a non-linear model such as mine with correlated inputs.

I have developed scripts for generating the model inputs and processing the outputs, and am now looking to see if there the methods apply...maybe I got that backwards...

spizwhiz:
I just wanted to briefly revisit this in case anyone is looking for this information.

According to Borgonovo(2007) the delta moment independent measure is valid even when parameters are correlated.  It seems to me then that the delta methods implemented here (from Plischke et al. (2013)) should also work for correlated parameter sets (and can even work with real data sets that are not sampled conditionally).

Feel free to let me know if you think I've mis-read something.

lukas1994:
Any updates on this? Would any of the implemented methods work if I provide my own sample (sample directly from the joint distribution)?

willu47:
> I just wanted to briefly revisit this in case anyone is looking for this information.
> 
> According to Borgonovo (2007) the delta moment independent measure is valid even when parameters are correlated. It seems to me then that the delta methods implemented in SALib (from Plischke et al. (2013)) should also work for correlated parameter sets (and can even work with real data sets that are not sampled conditionally). For an example, see Borgonovo and Tarantola (2008).


[...]

PR #275, opened 2019-11-08
HDMR

sahin-abdullah:
Added High Dimensional Model Representation (HDMR) introduced by [Li and Rabitz](https://pubs.acs.org/doi/pdf/10.1021/jp9096919) to the literature and five exact examples in their paper.

The distinguish feature of this method is that it provides an emulator as it decomposes models into its components. This emulator can be tested with different set of input matrix. Since this is a sensitivity analysis library, I did not add emulator part, even if I computed it. I provided figures for emulator part only.


sahin-abdullah:
Something seems wrong with delta_svm. It is the only one failed.

jdherman:
Thanks very much @sahin-abdullah. The HDMR method gives two very nice advantages that are not currently part of SALib: the ability to deal with correlated inputs (#193 ), and to estimate higher than first-order sensitivity from given data.

Because this is a larger contribution that will probably be used a lot, I have some questions/comments related to both the code and the method itself.

**Questions about the method**:

1. This method works in two parts: fitting a response surface to the data, then computing sensitivity indices that account for correlated inputs. The authors use two different names for these, HDMR and SCSA (structural and correlative sensitivity analysis). Are these two things separable, or are they required to go together? 

2. Something I've wondered about this method, the sensitivity results do not seem to include or compensate for the effect of the error of the response surface model. Should they? At least I think the user should have a clear idea of how accurate the model fit is.

**Comments on code**:

3. General style:
- Please use docstring format for the functions [(Example)](https://github.com/SALib/SALib/blob/master/src/SALib/analyze/sobol.py#L27)
- We are following PEP8 style as much as possible. This can be implemented using the [autopep8](https://pypi.org/project/autopep8/) tool.

4. Interpolation: most of the code deals with the cubic spline interpolation. I'm wondering if it would be possible to use built-in functions from either scipy or scikit-learn instead [(example)](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.CubicSpline.html#scipy.interpolate.CubicSpline). This would make the code cleaner and less prone to bugs in the future. I believe it is possible to recover the coefficients of the interpolation function after fitting, which could then be used to compute the sensitivity indices.

5. It's great that figures are included. These can go under the `plotting` submodule, so the usage would look something like this:
```python
from SALib.analyze import hdmr
from SALib.plotting import hdmr_plot  # (example)
# ...
result = hdmr(...)
hdmr_plot(result, ...)
```
I would also suggest one plotting function per figure type, rather than having them all appear at once.

6. Examples:
- It's not necessary to store the input/output txt files in the repository - these can be generated when running the example code.
- There probably only needs to be one example script for each method. In this case, the importance of the other case studies 1-5 is to confirm the results from the paper - so they would work better as tests (see below).

7. Tests:
- The case study examples can be implemented as regression tests, as you've already done for the Ishigami function.
- All of the assertions in the `hdmr_init` function are great, but please also add tests for these bad input cases [(example)](https://github.com/SALib/SALib/blob/master/tests/test_sobol.py#L69). 
- The assert statements should raise built-in Python error types, such as `RuntimeError` or `ValueError`, etc. [(example)](https://github.com/SALib/SALib/blob/master/src/SALib/analyze/sobol.py#L90).
- No warnings needed when assuming the default parameter settings - this should already be clear from the docstrings.

Thanks again for the effort to write up this method. It will be a great addition to the library.

sahin-abdullah:
You are welcome Professor. I am sorry I have not worked on modifications that you listed above. I have been busy these days. I will make those changes later. Happy Thanksgiving!

sahin-abdullah:
1. Yes, they are two separate things: (1) HDMR is a method to represent the model as the sum of its summands (or component functions). Unlike what Sobol introduced in his [study](https://pdfs.semanticscholar.org/d339/b9cc42d6a7286d96814e6713fd13cdde87e7.pdf) (variance decomposition), HDMR provides a functional decomposition of a model. (2) SCSA is a technique for covariance decomposition. When input variables are correlated, their corresponding component functions are also correlated. SCSA distinguishes component functions uncorrelated and correlated contributions.

2. The theory of HDMR goes back to a [paper](https://projecteuclid.org/euclid.aoms/1177730196) publish in 1948 by Hoeffding. He brought a term called u-statistics that accounts for unbiased estimation (unbiased variance, mean, etc.). How HDMR relates to u-statistics is that HDMR starts with subtracting the mean of Y from Y so that all remaining component functions have zero expected value. From that point on, HDMR determines component functions with the help of proper basis functions (b-splines, orthonormal polynomials) in an unbiased way. If there is an error in the model output, HDMR tries to fit that error as well. If the error is homoscedastic, it should not affect the sensitivity indices because of the equation below.


If the error is heteroscedastic and dominates the model output, that means HDMR could not decompose the model well and uncertainty in the input domain could not propagate to the model output well. In other words, there is no significant input-output relation. I attached two figures: (1) with zero mean normally distributed random error (as described in Li's paper) and (2) heteroscedastic error. 

Fig1:
homo_err

Fig2:
hetero_err

3. I used level 1 aggressiveness

4. Interpolation: There might be a misunderstanding with cubic splines from my side. I provide what part of the cubic spline function that I did not understand here. HDMR uses b-splines at the order of 3, not cubic splines. As far as I understand cubic spline function under scipy does interpolation at given query points. Also, it requires sorted x inputs. Our input, X, is not sorted. If we sort it, we distort the IO relation as well because certain input parameter is supposed to correspond to a certain output.  

5. We have better plotting function what it was before.

6. Done. But could not test with TravisCI yet.

7. Done.

Thank you for your comment on the code. I am here to edit it more if further feedback comes.

src/SALib/analyze/hdmr.py
```diff
@@ -0,0 +1,692 @@
...
+import time
+import numpy as np
+from SALib.plotting import hdmr
+from scipy import (stats, special, interpolate)
+from . import common_args
+from ..util import read_param_file, ResultDict
+from matplotlib import pyplot as plt
+
+
+def analyze(problem, X, Y, options={}):
```
willu47: Please follow numpydoc standards like this:

```python
def function(argument1, argument2=None):
    """One line description of the function

    Empty line above, and now this can be a longer paragraph describing the function

    Arguments
    ---------
    argument1 : int
        Description of argument1
    argument : float, default=None
        Description of argument2

    Returns
    -------
    float
        Description of the return value and type

    Notes
    -----
    Other notes etc.  For more information, see https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

File: tests/test_regression.py [hunk 2]
```diff
@@ -230,6 +233,94 @@ def test_regression_fast():
...
+            'maxiter': 100,
+            'm': 2,
+            'K': 1,
+            'R': 10000,
+            'alfa': 0.95,
+            'lambdax': 0.01,
+            'print_to_console': 0
+        } 
+    Si = hdmr.analyze(problem, X, Y, options)
+    assert_allclose(Si['Sa'][0:problem['num_vars']], [0.28, 0,17, 0.10, 0.04, 0.02], atol=5e-2, rtol=1e-1)
```
willu47: Look like you are missing a decimal point here - `0.17` which is causing the test to fail.

ConnectedSystems:
Hi @willu47

This PR will be ready for final review (if you feel it is warranted) after a final set of changes is merged in (see: https://github.com/sahin-abdullah/SALib/pull/3).

@jdherman and @willu47 I wonder if it would be possible to incorporate this into the v1.4 timeline - even if a warning is thrown stating that this is a tentative release and we're aiming for "official" inclusion in v1.5. To clarify I believe this implementation is robust, but there may be some confusion around the increased number of plots produced. Adding these features/methods earlier might help catch any issues that users find.

Ditto for the work I am currently doing described in https://github.com/SALib/SALib/issues/234#issuecomment-638741161 and maybe https://github.com/SALib/SALib/pull/265.

Thanks to @sahin-abdullah for this substantial piece of work.

PR #276, opened 2019-11-08
added tolerance to delta singular value matrix error test

ConnectedSystems:
Relaxed test conditions to address failing test for #275


ConnectedSystems:
Not sure what's changed recently to cause the test to fail, but I shouldn't have compared against exact values in any case.

jdherman:
Thanks @ConnectedSystems. I think the new regression test in #275 changed the sequence of random samples enough for the delta_svm test to fall outside the tolerance. Nothing changed with the delta method itself.

PR #280, opened 2020-01-17
Numpy >1.18 compatibility

bsteubing:
A "TypeError: object of type <class 'numpy.float64'> cannot be safely interpreted as an integer." is thrown by numpy >1.18 in the linspace function in the delta analysis. `int(round(M))` should fix this issue.

PR #285, opened 2020-03-06
Update travis and package config

ConnectedSystems:
Allow failure on Py3.5 - build is failing due to the use of features only found in Py3.6

Use of `python setup.py test` raises deprecation warning. Switched to `pytest` to resolve.

Updated documentation to match


ConnectedSystems:
I guess this is the first step towards formally deprecating support for Py3.5.

See also the comments relating to PR #280

PR #294, opened 2020-04-11
Identical default num_resamples for delta, dgsm and morris #291

### LINKED ISSUE #291: Why is the default value of num_resamples for Delta-MIM lower than for other methods?
Hi all,

Default num_resamples values for different SALib analyze methods:
Morris: num_resamples=1000
DGSM: num_resamples=1000
Sobol: num_resamples=100
Delta-MIM: num_resamples=10

Why is the default value for resamples lower for Delta-MIM (and Sobol) than Morris and DGSM?

And thanks for all the great work!

TobiasKAndersen:
Change default num_resamples values for Delta, DGSM and Morris to identical default num_resamples of 100 samples, which Closes #291

### DIFF
--- a/src/SALib/analyze/delta.py
+++ b/src/SALib/analyze/delta.py
@@ -9,7 +9,7 @@ from . import common_args
 from ..util import read_param_file, ResultDict
 
 
-def analyze(problem, X, Y, num_resamples=10,
+def analyze(problem, X, Y, num_resamples=100,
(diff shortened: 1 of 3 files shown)


jdherman:
Thanks @TobiasKAndersen !

TobiasKAndersen:
Of course! And thanks for all your great work you are doing on SALib

Issue #320, opened 2020-06-17
Enhancement proposal: code refactoring


lbteixeira:
Hello, guys.

I was checking this project today, and I found it really interesting. I've been making some sensitivity analysis during my PhD studies, so this project is really useful for me.

Checking the code, I realized that perhaps I could contribute someway. I would like to propose you a refactor of the code to make it cleaner and more readable. Here are some things I could do:

1. Change some variables' names to increase readability
2. Make sure the code is using PEP-8 style (for example, 79 characters per line, naming conventions, etc)
3. Abstract functions from the existing ones. The idea is to increase the test coverage and to increase readability, writing small functions that do only one thing
4. Write extra unit tests
5. Eliminate hard coding, creating new default variables for the functions (similar to what I did in issue #316 today)
6. Add annotations to the variables showing the expected types

My idea is to start small and to do only one change first. Then I would submit it and wait for your feedback.

How do you feel about that?

willu47:
Hi @lbteixeira. Many thanks for the suggestion. And thanks for the initiative and enthusiasm! I think these points would make a nice addition to the `CONTRIBUTING.md` guidelines.  We do not currently have a code-style section, and writing up a clear set of guidelines for contributors would be useful for the SALib community.

Code style can be contentious. For example, 79 characters is rather limiting from my point of view. Also, many of these aspects can be automated, and we are already using git hooks via the `pre-commit` to enforce certain aspects of code style.  And please, no automated-linting pull requests that reformat the entire code base...

Increasing test coverage is welcome as there are some parts of the package which are currently only covered by regression tests. So digging into the individual methods and targeted refactoring would be good. Ideally the tests should be useful to those reading them, explaining what the code should do, so examples from text books translated to the code are really handy.

lbteixeira:
Hello, @willu47. Thanks for your reply.

I agree 79 characters isn't always nice, and this isn't a real concern I have while writing code, although some developers get crazy if you don't follow this aspect of PEP-8. I'm happy to know you are more flexible about this. Anyway, the parts of the code I checked so far are almost entirely conforming to this 79 characters standard, so this isn't a main issue. And for sure, no automated-linting...

I'll suggest you that I start by working at the subpackage sample/morris, as this is more connected to what I did so far during my PhD. I'll make as many commits as I can, so you can track in details what I did and see what you like or not. Once we agree on the standards to follow, I can update the `CONTRIBUTING.md` file.

[...]

PR #322, opened 2020-06-19
Refactor Morris Issue 320 - Part 1

lbteixeira:
Hello, guys.

This is Part 1 of the refactoring I proposed in #320. There is more I would like to do, but I guess this is enough for a single PR. Probably the best idea would be to break #320 in more issues, to become easier to review the code.

My main modifications:
1. Excluded method `_sample_oat`, since `_sample_groups` is general enough to handle the task
1.1. Now, if the user doesn't define  `problem['groups']` in the input dictionary, this key takes the value of the variables names. In summary, everything works as there were no groups
1.2. If the user define the wrong number of entries in `problem['groups']`, the code continues to raise an exception
2. Renamed `_sample_groups` to `_sample_morris`
3. Implemented new functions to break the code into smaller pieces. Also implemented their test functions
4. Added some docstrings and modified some variables names to improve readability
5. Included some type hints

For more details, please check the commits history


src/SALib/sample/morris/__init__.py
```diff
@@ -142,7 +117,7 @@ def _sample_groups(problem, N, num_levels=4):
     ---------
```
ConnectedSystems, line 117: Let's standardize to `Parameters` rather than `Arguments`

src/SALib/sample/morris/__init__.py
```diff
@@ -346,6 +330,59 @@ def _compute_optimised_trajectories(problem, input_sample, N, k_choices,
...
+    ----------
+    num_levels: int
+        Number of levels
+    """
+    if not num_levels % 2 == 0:
+        warnings.warn("num_levels should be an even number, "
+                      "sample may be biased")
+
+
+def check_group_membership(group_membership):
```
ConnectedSystems, line 371: Annotating with `np.ndarray` or `np.array` is fine for our purposes.

```python
import numpy as np

def some_function(X: np.array) -> np.array:
    return X

File: tests/sample/morris/test_morris.py [hunk 3]
```diff
@@ -13,9 +13,12 @@
                                  compute_b_star,
                                  compute_delta,
                                  generate_trajectory,
-                                 generate_x_star)
+                                 generate_x_star,)
 from SALib.util import read_param_file, compute_groups_matrix
 
+from src.SALib.sample.morris import define_problem_with_groups, \
+    check_if_num_levels_is_even, check_group_membership
+
```
ConnectedSystems, line 21: Shouldn't need to specify `src` as the top-level module:

```python
from SALib.sample.morris import define_problem_with_groups

# vs

from src.SALib.sample.morris import define_problem_with_groups
```

I prefer using parentheses to group multi-line imports as it looks cleaner than using `\` to me.

```python
from SALib.sample.morris import (define_problem_with_groups,
                                 check_if_num_levels_is_even, 
                                 check_group_membership)
```

tests/sample/morris/test_morris.py
```diff
@@ -224,3 +222,105 @@ def test_generate_x_star(self):
...
+    def test_check_group_membership_value_error(self):
+        """
+        Checks if an error is raised if group_membership has None type.
+        """
+        # Creates a dummy variable
+        group_membership = None
+
+        with raises(ValueError):
+            assert check_group_membership(group_membership)
```
ConnectedSystems, line 326: Add newline at end of file to conform to PEP8.

These tests are excellent and worthwhile additions, thanks!

tests/sample/morris/test_morris.py
```diff
@@ -224,3 +222,105 @@ def test_generate_x_star(self):
...
+
+    def test_check_group_membership_type_error(self):
+        """
+        Checks if an error is raised if group_membership has a wrong type.
+        """
+        # Creates a dummy variable
+        group_membership = 1.0
+
+        with raises(TypeError):
+            assert check_group_membership(group_membership)
```
ConnectedSystems, line 316: Can this test be merged with the one above?

Specifically, `test_check_group_membership_all_ok()`

src/SALib/sample/morris/__init__.py
```diff
@@ -346,6 +330,59 @@ def _compute_optimised_trajectories(problem, input_sample, N, k_choices,
...
 
 
+def define_problem_with_groups(problem: dict) -> dict:
+    """
+    Checks if the user defined the 'groups' key in the problem dictionary.
+    If not, makes the 'groups' key equal to the variables names. In other
+    words, the number of groups will be equal to the number of variables, which
+    is equivalent to don't use groups at all.
+    Parameters
+    ----------
```
ConnectedSystems, line 340: Change awkward phrasing:

"is equivalent to don't use groups at all."

to

"is equivalent to no groups" (or "groupings"? - I'll leave you to decide)


Additionally, insert a newline between docstring text and parameter list for all functions, e.g:

```
[...] is equivalent to don't use groups at all.

Parameters
----------
problem : dict
```

src/SALib/sample/morris/__init__.py
```diff
@@ -51,8 +51,8 @@
 __all__ = ['sample']
 
 
-def sample(problem, N, num_levels=4, optimal_trajectories=None,
-           local_optimization=True, seed=None):
+def sample(problem, number_trajectories, num_levels=4,
```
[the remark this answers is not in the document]
lbteixeira: I agree that changing the variable name in the function `sample`can be misleading, I just reverted back. I annotated the `seed`parameter as `int`, but I'm not sure this is the best choice. Feel free to ask me to change if you want

tests/sample/morris/test_morris.py
lbteixeira: Normally I like to have only one assert statement per test, but no problem. I'll merge the two last tests, but `test_check_group_membership_all_ok` is marked as expected to fail.

tests/sample/morris/test_morris.py
lbteixeira: For some strange reason, when I remove the `src` I get an import error. This is only happening with the functions I created, I don't know exactly why. I tried to google this error, but didn't find a good answer. Can you take a look, please? For now I just left the code as it was.

ConnectedSystems: Sure, I will do that asap

ConnectedSystems: In `__init__.py` you have to import the high-level functions you wish users to see.

```python
# in __init__.py

from .morris import sample
```

This will allow users to import Morris sampling as before.

```python

from SALib.sample.morris import sample as morris_sample
```

As you want to import specific functions in the context of testing, you have to specify the file itself in the import statement:

```python

# import from the sample dir -> morris dir -> morris file
from SALib.sample.morris.morris import (define_problem_with_groups,
                                        check_if_num_levels_is_even, 
                                        check_group_membership)
```

Alternatively, you could specify those functions in the morris package `__init__.py` as well but this will pollute the namespace and auto-complete, allowing users who don't usually have a reason to be aware of these functions to see them.

lbteixeira: Hello, @ConnectedSystems. Thanks for your reply. Being relatively new to Python, I'm trying to figure out how exactly all those import statements work. I would like to share what I have seen so far and ask for your opinion (I'm sorry if this reply is too large).

In the module `morris`, inside the subpackage `morris`, module `common_args` is imported using the following command: `from SALib.sample import common_args`.

When I hover my mouse over `SALib`, Pycharm shows me the following warning:

image

Indeed, this "Package containing module SALib" isn't listed in the requirements,  because that is the package SALib itself. At first I considered this a little strange, because I had to install SALib in my virtual environment to develop SALib. Since all files are in the same project, I believe it shouldn't be necessary to do that, but if I don't, I get an error.

The solution to this would be to import using the folders hierarchy. Since the package `sample` is one level up the package `morris`, this import would become:

`from .. import common_args`

This works and gives no warnings. The same holds to import modules from another package, and this is were this discussion started. 

In my code, inside the `tests` package, I used this: `from src.SALib.sample.morris.morris import (check_group_membership)`. What I'm doing here is to make an absolute import _inside_ the project (as suggested by Pycharm), with no dependencies from any package installed in my virtual env.

In this same package there is another import: `from SALib.sample.morris import (sample)`. This is importing from _outside_ the project, i. e., it's importing from the package SALib installed in my virtual env, not the one I'm currently developing. The downside of this is that, if I make local changes to my modules, this import won't be able to "see" this.

As far as I understand, this is exactly what happened here. I added three extra functions to the package `morris`, but I couldn't import them using just `from SALib.sample.morris.morris` because the installation in my virtual env is now outdated. I would have to install it again, this time with the functions I wrote included (or use another technique to automatically update it, I believe this can be done).

Finally, if everything I wrote is correct, I believe the best option would be:

1. Replace all import statements that use `SALib.<something>` by the correspondent import using the folders hierarchy.
2. Continue to use absolute imports in the package `tests`
3. Expose the method `sample` to the user using the `__init__.py` file, as you suggested

I hope I was clear, and I'm looking forward for your reply. Don't hesitate to ask me if you didn't understand something.

A good reference can be found here: https://realpython.com/lessons/subpackages/

ConnectedSystems: Hi,

Not sure I followed everything, I'll have another read through later when my mind is clearer.

One thing that's clear though is that you have multiple copies of SALib floating around.
In the vast majority of cases you shouldn't have to do this.

Set your project up so that you have one (and only one) project folder which you are working on and install the project in "development mode" (or an "editable install"). 

From the project root (top of project directory):

```bash
# Yes the period is important
$ pip install -e .

# Or if using conda and conda-build
$ conda develop .

# Or if you have neither of the above
$ python setup.py develop
```

See here for more explanation:
https://stackoverflow.com/a/61813942/2694952

Doing the above means you don't have to copy files back and forth which is a source of issues when working with many people and branches.

lbteixeira: Hi, @ConnectedSystems. Thanks for this very interesting info about the develop mode, I wasn't aware about it. 

Actually, I only have one copy of SALib source code. What I was saying is that I installed SALib using regular `pip install salib` directly from PyPI in a virtual environment.

As I understand, there are basically 4 options:

1. Regular `pip install salib` directly from PyPI (as I did). **If I need to have SALib installed to run the tests**, this is a terrible choice, since what I change in the code doesn't affect my installed version, of course. My point is: in my opinion, I shouldn't need to have the package actually installed in my virtual env to run some piece of test code. Since the tests are in the same folder as my source code (but not in the same package), the changes I do should be visible to them.

2. Don't install SALib at all. The way the code is today, this wouldn't work. As I said in my last answer, there are some imports that require SALib to be installed, such as `from SALib.sample import common_args`

3. Install SALib in development mode. This works, and actually fix all the problems. But SALib continues to make imports from itself, what I think it's weird (again, this is just my opinion...)

4. Install SALib in development mode **AND** remove the dependency from SALib itself, as I said in the previous reply. By doing this, SALib won't require itself to be installed to work, and at the same time the new features included would be ready to be used in this virtual env.

Please, let me know what you think about that, I hope I was more clear this time. Nevertheless, I believe it's a little difficult to explain this without showing my screen, so I'm available to a quick video call if you want.

ConnectedSystems: Hi again,

A package referencing itself is not that big a deal although I understand your struggle to reconcile the conceptual mismatch. The end result is the same regardless of how the `import` statement is structured as Python resolves the statement to the same file and code.

Given we're developing a package it makes sense for the package to be installed (albeit in development mode).

Regarding the need to have the package installed to run tests: Tests should be run with the package installed. We're not, in this context, testing a loose collection of scripts - we are testing the correct functioning of a package.

Lastly, PEP8 recommends favouring absolute imports with relative imports acceptable when dealing with a complex package layout (https://www.python.org/dev/peps/pep-0008/#imports).

The example given...

```python
import mypkg.sibling
from mypkg import sibling
from mypkg.sibling import example
```

... lines up with how SALib, [and other Python packages](https://github.com/pandas-dev/pandas/blob/master/pandas/core/api.py), approach importing own code.

Hope this clears things up.

lbteixeira: Hi, @ConnectedSystems.

Thank you for your answer, everything is clear now. I'm happy to be contributing here, I'm learning a lot (I guess this is also the point of contributing to open source projects).

I'll commit my branch asap.

tests/sample/morris/test_morris.py
```diff
@@ -7,19 +7,18 @@
...
+from SALib.sample.morris import (sample)
+from SALib.sample.morris.morris import (check_group_membership,
+                                        check_if_num_levels_is_even,
+                                        define_problem_with_groups,
+                                        _compute_optimised_trajectories,
+                                        generate_p_star,
+                                        compute_b_star,
+                                        compute_delta,
+                                        generate_trajectory,
+                                        generate_x_star,)
```
ConnectedSystems, line 19: In this specific case we can group these two import statements together:

```
from SALib.sample.morris.morris import (sample,
                                        check_group_membership,
                                        check_if_num_levels_is_even,
                                        define_problem_with_groups,
                                        _compute_optimised_trajectories,
                                        generate_p_star,
                                        compute_b_star,
                                        compute_delta,
                                        generate_trajectory,
                                        generate_x_star)
```

ConnectedSystems:
Hi @lbteixeira (pinging @willu47) after the last comment (re combining import statements) is addressed I think this is ready to be merged.

willu47:
Thanks @ConnectedSystems and @lbteixeira - I've fixed the conflict between master and this branch which was stalling the CI from running. If the tests pass we can merge this in.

ConnectedSystems:
I'll address the failing tests - just a minor change to adjust existing tests to use the new morris functions

Issue #325, opened 2020-07-06
Refactoring Morris sample module


lbteixeira:
Hello. I'll propose a few more changes on the sampling module `morris.py`. No functionalities will be changed for now, I'll just try to clean the code a little more. Specially, I'll break the function `_compute_optimised_trajectories` in more functions.

willu47:
Hi @lbteixeira - Another thing which need doing is removing the brute force optimised trajectories code which depended upon Gurobi. This has been superseded by the locally optimal solution. Alternatively, it could be implemented in an open-source optimisation framework such as PuLP or pyomo, although those are both large dependencies which may not be worth it.

lbteixeira:
Hello, @willu47. Ok, I'll delete the brute force method. Later, if we find it's necessary to have it, we can implement it using these frameworks you cited.

willu47:
I actually meant the `gurobi.py` file - the brute force method is impractical for normal use, but it for very small tests to compare the other methods against.  Suggest you retain brute force and remove the gurobi file (and all other gurobi-related references).

lbteixeira:
Oh, ok. Sorry about that. I'll do as you said.

PR #326, opened 2020-07-06
Issue 325

lbteixeira:
Hello.

I've made a few changes to `morris.py` to make it a little more readable, as I said in #325. As suggested by @willu47, I removed everything related to `gurobi.py`.

One point: in the function `_compute_optimised_trajectories` an error is raised if `local_optimization` is false and `k_choices > 10`. Shouldn't this be a warning? Anyway, I didn't change this for the moment.


src/SALib/sample/morris/morris.py
```diff
@@ -303,42 +300,27 @@ def _compute_optimised_trajectories(problem: dict, input_sample: int, N: int,
...
-    output = context.sample(input_sample, N, num_params,
-                            k_choices, num_groups)
+
+    output = context.sample(input_sample, N, num_params, k_choices, num_groups)
 
     return output
 
 
-def define_problem_with_groups(problem: dict) -> dict:
+def _define_problem_with_groups(problem: dict) -> dict:
```
willu47, line 323: Use `from typing import Dict` instead of `dict`

src/SALib/sample/morris/morris.py
```diff
@@ -132,19 +124,24 @@ def _sample_morris(problem: dict, number_trajectories: int,
...
-    sample = np.array([generate_trajectory(group_membership,
-                                           num_levels)
-                       for n in range(number_trajectories)])
-    return sample.reshape((number_trajectories * (num_groups + 1), num_params))
+    i = 0
+    sample_morris = []
+    while i < number_trajectories:
+        sample_morris.append(_generate_trajectory(group_membership, num_levels))
+        i += 1
+    sample_morris = np.array(sample_morris)
```
willu47, line 137: While this is more readable than what it replaces, I wonder if it is necessary to introduce a counter.  `for n in range(number_trajectories)` seems sufficiently clear.  I would tidy this as follows:
```python
sample = [generate_trajectory(group_membership, num_levels) for n in range(number_trajectories)]
sample_morris = np.array(sample).reshape((number_trajectories * (num_groups + 1), num_params))
return sample_morris
```
But perhaps this is just bike-shedding...

src/SALib/sample/morris/morris.py
```diff
@@ -303,42 +300,27 @@ def _compute_optimised_trajectories(problem: dict, input_sample: int, N: int,
...
-        strategy = LocalOptimisation()
-    else:
-        # Use brute force approach
-        strategy = BruteForce()
-
-    if problem.get('groups'):
-        num_groups = len(set(problem['groups']))
-    else:
-        num_groups = None
+    num_groups = len(set(problem['groups']))
```
willu47, line 312: Groups are now always present so we don't need this check?

src/SALib/sample/morris/morris.py
```diff
@@ -379,7 +361,7 @@ def check_if_num_levels_is_even(num_levels: int):
                       "sample may be biased")
 
 
-def check_group_membership(group_membership):
+def _check_group_membership(group_membership):
```
willu47, line 364: typing for argument?

src/SALib/sample/morris/morris.py
```diff
@@ -85,28 +77,28 @@ def sample(problem: dict, N: int, num_levels: int = 4,
...
+    sample_morris : numpy.ndarray
         Returns a numpy.ndarray containing the model inputs required for Method
         of Morris. The resulting matrix has :math:`(G/D+1)*N/T` rows and
         :math:`D` columns, where :math:`D` is the number of parameters.
     """
     if seed:
         np.random.seed(seed)
 
-    check_if_num_levels_is_even(num_levels)
+    _check_if_num_levels_is_even(num_levels)
```
willu47, line 88: typing?

src/SALib/sample/morris/morris.py
lbteixeira: My "problem" with this approach is that `n` isn't actually used for anything, that's why I prefer the `while` over the `for` loop, even if I have to include a counter. Besides, I have the impression that using `reshape` makes the code more difficult to understand. It's up to you, though...

src/SALib/sample/morris/morris.py
lbteixeira: That actually came with my last contribution. My commit description wasn't really accurate, because I wanted to be short. But, what happens in the code now is that, if the user doesn't define groups, the function `_define_problem_with_groups` makes the `groups` key equal to the variables names. In other words, the number of groups will be equal to the number of variables, which is equivalent to no groups. This eliminates some `if` statements, including this one, and also some extra code.

src/SALib/sample/morris/morris.py
lbteixeira: This is actually a function call, not a definition. There is a `num_levels: int` in the function definition.

src/SALib/sample/morris/morris.py
ConnectedSystems: Using `range()` is more pythonic in this case.

If the value (`n`) is not used then Python convention is to indicate that with an underscore (`_`), for example:

`[generate_trajectory(group_membership, num_levels) for _ in range(number_trajectories)]`

See [here](https://stackoverflow.com/a/47599668/2694952) for another use of the underscore.

lbteixeira: Great, @ConnectedSystems. I just made the change as you proposed.

src/SALib/sample/morris/morris.py
willu47: Oops, you are correct!

lbteixeira:
Is this ready to be merged? I depend on it to work on #327 :)

ConnectedSystems:
Hi @lbteixeira 

No need to wait for this to be merged - simply create a new branch off this branch (`issue_325`) and continue working as normal.

lbteixeira:
Hi, @ConnectedSystems. I was afraid to create conflicts if later I had to change something at this branch (`issue_325`), but I think no important problems will be created.

Issue #329, opened 2020-07-09
Sample problem in boundary-distribution definition: bounds are not legal.


MatthiVH:
The saltelli sampler can handle other distributions in the problem definition, however the latin hypercube sampler can not.

Example-code:
```python
problem = { 'num_vars': 3,
                    'names': ['P1', 'P2', 'P3'],
                    'bounds': [[0.9, 1.1], [5, 1], [3.4, 0.1]],
                    'dists': ['unif', 'norm', 'lognorm']}

#generating samples
saltelli_samples = saltelli.sample(prob_dists_code, 1000, calc_second_order=True) -> works fine

#generating samples
latin_samples = latin.sample(prob_dists_code, 1000) -> ValueError: Bounds are not legal.
```

I assume the latin sampler doesn't recognize the 'dists'-statement in the problem-definition and assumes all uniform.
Can this be fixed?

Regards,
Matthias

willu47:
Hi @MatthiVH - thanks for the bug report. We'll investigate. It should be a relatively easy fix (I hope).

MatthiVH:
Thanks, that's really helpful.

ConnectedSystems:
Addressed by PR for v1.4 so I will close this.

Thanks again @lbteixeira

PR #330, opened 2020-07-10
Latin sampler now works with different distributions

lbteixeira:
Code to solve issue #329.


willu47:
Thanks for the quick fix @lbteixeira!

Issue #332, opened 2020-07-13
Documentation ReadTheDocs


rddaz2013:
First ... Nice Work thx

On Page : https://salib.readthedocs.io/en/latest/basics.html

The ghosts of "Python 2" appear again in the print expression in the example

best wishes

ConnectedSystems:
Thanks for raising the issue. I will be cleaning up the documentation in the near future.

ConnectedSystems:
This is addressed in #346 so am closing now. Changes will be available in v1.4

Issue #308, opened 2020-04-26
Improve documentation of input distributions


willu47:
As raised by #307, we should add better documentation of the distributions you can select for inputs to a problem.

Issue #343, opened 2020-09-09
Morris sample problem in boundary-distribution definition: bounds are not legal. #329


MatthiVH:
Same issue as #329 but now for the Morris sampler.

The saltelli sampler can handle other distributions in the problem definition, however the Morris sampler can not.

I assume the Morris sampler doesn't recognise the 'dists'-statement in the problem-definition and assumes all uniform.
Can this be fixed?

Regards,
Matthias

ConnectedSystems:
I'll have a look at this sometime over the next few days - finding time is an issue at the moment so I can't promise anything concrete.

ConnectedSystems:
Hi @MatthiVH 

I'm not sure when the next release will be made, but in the meantime you can try out the "bleeding edge" development version:

Please do note that this version has a number of changes and has not been extensively tested.

```bash
pip install git+https://github.com/SALib/SALib.git@experimental-methods
```

Specific to your case, sampling non-uniform distributions may not work with groups of parameters (for Morris in particular).

PR #192, opened 2018-02-02
Feature - non uniform sampling - DNM

willu47:
Replaces #175 (see for discussion on remaining work)

- Rebased on latest version of code
- Fixed erroring tests

PR #346, opened 2020-09-12
Allow alternate sampling distributions for all methods

### LINKED ISSUE #329: Sample problem in boundary-distribution definition: bounds are not legal.

ConnectedSystems:
Generalize scaling process for all sampling methods.

Note: The `unscale_samples` will need to be adjusted or generalized but it does not appear to be used internally.

@willu47 Would you be available to push out a new release once I add documentation?

Replaces #192 (with apologies to @willu47 and @filssavi)
Resolves #329 #343 

pinging @MatthiVH


ConnectedSystems:
Added documentation for sampling non-uniform distributions

Resolves #308 

Partially addresses #274

pinging @DmitriyValetov regarding #338

ConnectedSystems:
In #339, prototype unified interface is provided, with the intention that sample scaling could be conducted at a single place in the codebase. Some methods, however, cannot be scaled post hoc (e.g. DGSM) so this needs to be kept in mind for the problem-spec approach, and possibly other aspects relating to #234.

This is one reason why use of the `apply_scaling()` function modifies the problem dict in place with a `samples_scaled` entry ([see here](https://github.com/SALib/SALib/pull/346/commits/47459fc2ad7baa521748ed45f95c046060cf940c#diff-36096cdb6888d81b6a7076038830c6c2R194)).

These specific cases will have to be handled...

ConnectedSystems:
Updated documentation will also resolve #332

PR #356, opened 2020-09-15
Fix out of order group names

ConnectedSystems:
Non-matching order results in NaNs reported for S2

### DIFF
--- a/src/SALib/analyze/sobol.py
+++ b/src/SALib/analyze/sobol.py
@@ -68,7 +68,7 @@ def analyze(problem, Y, calc_second_order=True, num_resamples=100,
     if not groups:
         D = problem['num_vars']
     else:
-        D = len(set(problem['groups']))
+        _, D = extract_group_names(groups)
 
     if calc_second_order and Y.size % (2 * D + 2) == 0:
         N = int(Y.size / (2 * D + 2))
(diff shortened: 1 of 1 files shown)


src/SALib/analyze/sobol.py
```diff
@@ -161,9 +160,9 @@ def create_Si_dict(D, calc_second_order):
                    for k in ('S1', 'S1_conf', 'ST', 'ST_conf'))
 
     if calc_second_order:
-        S['S2'] = np.zeros((D, D))
+        S['S2'] = np.empty((D, D))
```
willu47, line 163: @ConnectedSystems - I'm sure you're aware, but note we've had issues with using `np.empty` in the past because it doesn't necessarily return an array of zero values...  Unless immediately filling the array with values, suggest using `np.zeros`

ConnectedSystems: Hi Will,

Yes, this was a conscious decision. All array elements get immediately replaced with `np.nan` (see next line).

I replaced `np.zeros` with `np.empty` as the latter is much faster in producing an array of the desired size.
But perhaps this is an overzealous micro-optimization.

image

ConnectedSystems: I'll replace with `np.full()` as it's perhaps more readable

PR #358, opened 2020-09-18
Group analysis support

ConnectedSystems:
Support analysis of parameter groups, or raise warning when they are not supported.

DGSM and HDMR implementations were a little over my head so could not adjust the code to allow group sampling/analysis at this time (perhaps @sahin-abdullah would be willing to help here?).

### DIFF
--- a/src/SALib/analyze/delta.py
+++ b/src/SALib/analyze/delta.py
@@ -4,7 +4,10 @@ from scipy.stats import norm, gaussian_kde, rankdata
 import numpy as np
 
 from . import common_args
-from ..util import read_param_file, ResultDict
+from ..util import read_param_file, ResultDict, extract_groups
+
+import warnings
+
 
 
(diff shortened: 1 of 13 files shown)


lbteixeira:
Hello, @ConnectedSystems.

Sure, it would be great to help, but this week I'm a little too busy with my PhD thesis. I can work on this in about 1 week, if it's not too late.

ConnectedSystems:
Thanks, @lbteixeira  

There's no deadline and so. no rush. I'll come back to this myself in a week or so as well.

willu47:
One thing to consider with grouping approaches is that you can get cancellation effects. Morris uses an average of absolute elementary effect when computing sensitivity indices for groups, rather than the average. 

For example with two parameters in one group, if one parameter has a strong positive effect, and another a strong negative effect, these will cancel out when averaged and the group will seem to have a negligible effect.

PR #382, opened 2020-11-08
Fix minor type in RBD-FAST Tarantola reference hyperlink

Vaibhavdixit02:
Minor typo, noticed it while browsing the docs


ConnectedSystems:
Hi @Vaibhavdixit02 

Thanks for this.

I'll merge this into the main branch so the correction is immediately available.

Do note that we've already addressed this for v1.4 already but we're finalizing a few things though so I'm not sure when it will be released (see https://github.com/SALib/SALib/blob/develop/docs/index.rst)

Thanks again

Vaibhavdixit02:
>Do note that we've already addressed this for v1.4 already but we're finalizing a few things so I'm not sure when it will be released (see https://github.com/SALib/SALib/blob/develop/docs/index.rst)

Oh great, sorry didn't see that.

>I'll merge this into the main branch (when the checks pass) so the correction is immediately available.

Makes sense 👍

ConnectedSystems:
(ignore the close re-opening - attempting to restart travis-ci build as it appears to be stuck)

PR #395, opened 2021-01-14
Problem Spec setters

ConnectedSystems:
Minor enhancement for the SALib Problem class.

The effort to streamline the usage approach preferences a method chaining approach:

```python
(sp.sample(saltelli.sample, 1000)
   .evaluate(Ishigami.evaluate)
   .analyze_sobol())
```

Having samples/results available beforehand requires a procedural approach to be taken:

```python
sp.samples = previous_samples
sp.results = pre_existing_model_outputs
sp.analyze_sobol()
```

This commit provides convenience methods to allow the method chaining approach to be applied in either context:

```python
(sp.set_samples(previous_samples)
   .set_results(previous_results)
   .analyze_sobol())
```


ConnectedSystems:
CI fails on Py3.6 because Pandas (v1.2.1) fails to install for some reason with the message:

"RuntimeError: Cannot cythonize without Cython installed."

Not sure why as production versions of Pandas should not need Cython...

Any ideas @willu47 ?

PR #396, opened 2021-01-14
Add agentpy package to software list

JoelForamitti:
See here about how agentpy uses the SALib package: https://agentpy.readthedocs.io/en/latest/reference_analysis.html#agentpy.sensitivity_sobol


ConnectedSystems:
Thanks!

Issue #389, opened 2020-11-19
RuntimeWarning emitted in test for HDMR


ConnectedSystems:
RuntimeWarning from numpy. I am not sure how much impact it has or if it should be expected given the context.

```bash
tests/test_regression.py::test_regression_hdmr_ishigami
  /home/travis/miniconda/envs/testenv/lib/python3.7/site-packages/numpy/linalg/linalg.py:2159: RuntimeWarning: overflow encountered in det
    r = _umath_linalg.det(a, signature=signature)
```

Could be emitted if some subset of an array happens to be filled with constant values (e.g. during resampling?).

Possibly related issues:

https://github.com/numpy/numpy/issues/8529

roping @sahin-abdullah into this conversation

ConnectedSystems:
Fixed by #398

PR #398, opened 2021-01-20
NumPy Overflow warning when computing determinant

sahin-abdullah:
Regarding to issue #389. I was using determinant to test whether the matrix is full rank or not, however, this method creates an overflow problem. With the updated version, we check the ill-conditionality of the matrix with singular value decomposition.


ConnectedSystems:
Thanks!

PR #411, opened 2021-03-29
Acknowledge inspiration of eFAST implementation

ConnectedSystems:


willu47 (commented):
Having reviewed the text of the GPL v2 license and read a few articles about license compatibility, I believe that this strong copyleft license is incompatible with including the ported version of the fast code from R. The terms of the license state that software which includes the modified code must also be released under a GPL v2 license. This is not the case for a weak copyleft license, such as LGPL v3, which would only require us to release the modified code under the same license, with the remaining software under the existing MIT license.

As it is, we're technically in breach of the original license while we distribute this ported R code under an MIT license.

Solutions:

- replace the current implementation of FAST with an original implementation
- relicense SALib under a GPL v2 license

ConnectedSystems:
Thanks for the comments @willu47 

As discussed via email along with @jdherman I've since had a closer look/comparison with the original R code.
It is my opinion that we are not in breach given:
- the port is not a direct line-by-line translation of the original R code
- changes made since the original Python implementation makes the differences even more evident
- what similarities there are can be attributed to the nature of the algorithm itself rather than any implementation detail (algorithms not being subject to copyright or licensing issues)

Given the above, I am now in favour of adding an acknowledgement to the original R implementation and removing the GPL v2 license notice.

Would this resolve this issue @willu47 @jdherman ?

willu47:
This seems like a pragmatic way forward.

jdherman:
Thanks @ConnectedSystems, sorry for the delay. This update makes sense to me.

PR #413, opened 2021-04-09
support DataFrame conversion of sobol results when groups are defined…

menegon:
support DataFrame conversion of sobol results when groups are defined… into the problem


ConnectedSystems:
Hi @menegon 

Thanks for the contribution.

We have tentative support for this in the development branch, intended for v1.4 (see https://github.com/SALib/SALib/pull/392).
That said, I will review your suggested changes here in view of supporting v1.3.x

In the meantime you might be interested in trying out v1.4 (`git checkout develop`)

PR #416, opened 2021-04-19
Sobol sequence fix #414

### LINKED ISSUE #414: Wrong usage of Sobol' sequence
I noticed that Sobol' sequence could be used in a wrong way by `sample.saltelli`:

https://github.com/SALib/SALib/blob/6765749fcbb1d63a45a1021ef8061701aa2dc33d/src/SALib/sample/saltelli.py#L43

The convergence properties of Sobol' sequence are only valid if sampling exactly 2^n points. Also, if skipping is wanted, you must skip 2^m points and then you can only sample 2^n points with n < m.

We had lengthly discussions while adding a QMC module to SciPy notably with Art Owen, Sergei Kucherenko and Fred Hickernell:
https://github.com/scipy/scipy/pull/10844

Art Owen wrote an article about it too:
https://arxiv.org/abs/2008.08051

In the end, I would suggest to either fix the usage of Sobol' here or rely on the new `scipy.stats.qmc` which will be available in the next release of SciPy (1.7).

In general, this module if fairly new (and we are always looking for people interested in contributing) but already has Sobol' sequence, Halton, LHS and discrepancy measures. Implementations of Sobol' and discrepancy are in Cython and orders of magnitudes faster than equivalent in python.

Let me know if I can help you with something.

ps. Sobol' writes with the apostrophe, it's not a mark of possessive but the diacritic is in the name.

ConnectedSystems:
PR for Issue #414

@tupui - This PR raises a FutureWarning indicating that possible issues may arise in cases where the requirements are not met.
This is to allow older code to continue to function for now.

v1.4 onwards will instead raise a ValueError.

### DIFF
--- a/src/SALib/sample/saltelli.py
+++ b/src/SALib/sample/saltelli.py
@@ -13,11 +13,11 @@ def sample(problem, N, calc_second_order=True, seed=None, skip_values=1024):
     """Generates model inputs using Saltelli's extension of the Sobol' sequence.
 
     Returns a NumPy matrix containing the model inputs using Saltelli's sampling
-    scheme.  Saltelli's scheme extends the Sobol' sequence in a way to reduce
-    the error rates in the resulting sensitivity index calculations.  If
+    scheme. Saltelli's scheme extends the Sobol' sequence in a way to reduce
+    the error rates in the resulting sensitivity index calculations. If
     calc_second_order is False, the resulting matrix has N * (D + 2)
-    rows, where D is the number of parameters.  If calc_second_order is `True`,
-    the resulting matrix has N * (2D + 2) rows.  These model inputs are
+    rows, where D is the number of parameters. If calc_second_order is `True`,
+    the resulting matrix has N * (2D + 2) rows. These model inputs are
     intended to be used with :func:`SALib.analyze.sobol.analyze`.
 
     Parameters
@@ -31,7 +31,7 @@ def sample(problem, N, calc_second_order=True, seed=None, skip_values=1024):
     """
     if seed:
         msg = "The seed value is ignored for the Saltelli sampler\n"
-        msg += "as it uses the (deterministic) Sobol sequence.\n"
+        msg += "as it uses the (deterministic) Sobol' sequence.\n"
         msg += "Different samples can be obtained by setting the\n"
         msg += "`skip_values` parameter (defaults to 1024)."
         warnings.warn(msg)
--- a/src/SALib/sample/saltelli.py
+++ b/src/SALib/sample/saltelli.py
@@ -17,8 +17,9 @@ def sample(problem, N, calc_second_order=True, seed=None, skip_values=1024):
     the error rates in the resulting sensitivity index calculations. If
     `calc_second_order` is False, the resulting matrix has ``N * (D + 2)``
 
-    rows, where D is the number of parameters. If calc_second_order is `True`,
-    the resulting matrix has N * (2D + 2) rows. These model inputs are
+    rows, where ``D`` is the number of parameters. If `calc_second_order` is True,
+    the resulting matrix has ``N * (2D + 2)`` rows. These model inputs are
+
     intended to be used with :func:`SALib.analyze.sobol.analyze`.
 
     Parameters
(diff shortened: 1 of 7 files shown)


src/SALib/sample/saltelli.py
```diff
@@ -1,21 +1,23 @@
...
     Returns a NumPy matrix containing the model inputs using Saltelli's sampling
-    scheme.  Saltelli's scheme extends the Sobol sequence in a way to reduce
-    the error rates in the resulting sensitivity index calculations.  If
+    scheme. Saltelli's scheme extends the Sobol' sequence in a way to reduce
+    the error rates in the resulting sensitivity index calculations. If
     calc_second_order is False, the resulting matrix has N * (D + 2)
-    rows, where D is the number of parameters.  If calc_second_order is True,
-    the resulting matrix has N * (2D + 2) rows.  These model inputs are
+    rows, where D is the number of parameters. If calc_second_order is `True`,
+    the resulting matrix has N * (2D + 2) rows. These model inputs are
```
tupui, line 20:
```suggestion
    rows, where ``D`` is the number of parameters. If `calc_second_order` is True,
    the resulting matrix has ``N * (2D + 2)`` rows. These model inputs are

```

src/SALib/sample/saltelli.py
```diff
@@ -1,21 +1,23 @@
...
-    """Generates model inputs using Saltelli's extension of the Sobol sequence.
+def sample(problem, N, calc_second_order=True, seed=None, skip_values=1024):
+    """Generates model inputs using Saltelli's extension of the Sobol' sequence.
 
     Returns a NumPy matrix containing the model inputs using Saltelli's sampling
-    scheme.  Saltelli's scheme extends the Sobol sequence in a way to reduce
-    the error rates in the resulting sensitivity index calculations.  If
+    scheme. Saltelli's scheme extends the Sobol' sequence in a way to reduce
+    the error rates in the resulting sensitivity index calculations. If
     calc_second_order is False, the resulting matrix has N * (D + 2)
```
tupui, line 18:
```suggestion
    `calc_second_order` is False, the resulting matrix has ``N * (D + 2)``

```

src/SALib/sample/saltelli.py
```diff
@@ -28,7 +30,46 @@ def sample(problem, N, calc_second_order=True, seed=None, skip_values=1000):
         Calculate second-order sensitivities (default True)
     """
     if seed:
-        np.random.seed(seed)
+        msg = "The seed value is ignored for the Saltelli sampler\n"
+        msg += "as it uses the (deterministic) Sobol' sequence.\n"
+        msg += "Different samples can be obtained by setting the\n"
+        msg += "`skip_values` parameter (defaults to 1024)."
```
tupui, line 36:
```suggestion
        msg = ("The seed value is ignored for the Saltelli sampler\n"
               "as it uses the (deterministic) Sobol' sequence.\n"
               "Different samples can be obtained by setting the\n"
               "`skip_values` parameter (defaults to 1024).")

```

tupui (commented):
Thanks for fixing this! I made some stylistic suggestions.

Also, I don't know about your testing policy, but I would suggest to add a test to catch the warnings.

src/SALib/sample/saltelli.py
```diff
@@ -1,34 +1,79 @@
...
 import numpy as np
 
 from . import common_args
 from . import sobol_sequence
 from ..util import scale_samples, nonuniform_scale_samples, read_param_file, compute_groups_matrix
 
 
-def sample(problem, N, calc_second_order=True, seed=None, skip_values=1000):
-    """Generates model inputs using Saltelli's extension of the Sobol sequence.
+def sample(problem, N, calc_second_order=True, seed=None, skip_values=1024):
```
tupui, line 12: The seeding could also be removed. But might break some codes.
```suggestion
def sample(problem, N, calc_second_order=True, skip_values=1024):

```

ConnectedSystems: I'll keep `seed` for now, but might remove in the dev branch

ConnectedSystems:
> Thanks for fixing this! I made some stylistic suggestions.
> 
> Also, I don't know about your testing policy, but I would suggest to add a test to catch the warnings.

I've added targeted tests for v1.4 as I expect this version to be relatively short-lived

PR #418, opened 2021-04-28
Require setuptools, src/SALib/__init__.py, /setup.py and docs/conf.py import pkg_resources

hrnciar:


ConnectedSystems:
Hello,

Thanks for the PR.

Could you give a brief overview of what issue this PR addresses? I assume it is to enable local installation via pip or something?

hrnciar:
Hello, `pkg_resources` needs setuptools as a runtime dependency. Since it wasn't specified so far, there was a possibility of runtime failure in the case of setuptools not being present. There is a way how to install packages without setuptools or setuptools can be uninstalled after SALib package was installed. My PR prevents this error.

ConnectedSystems:
Thanks again for the PR.

It looks good to me. The tests fail due to some checks that rely on no warnings being raised. Warnings that were not previously raised are caused by changes to numpy (deprecations coming in numpy v1.2).

I have commited some changes and will merge once the tests are confirmed to pass on travis again.

Issue #422, opened 2021-05-31
[bug] Sobol range check for conf_level is incorrectly inclusive of 0 and 1


scottshambaugh:
conf_level should be within, but exclusive of 0 and 1. 1 leads to an infinite value later on in the line `Z = norm.ppf(0.5 + conf_level / 2)`, and 0 give a nonsensical result. Opening a PR here shortly.

ConnectedSystems:
Closing as PR has been merged. Thanks again for catching this.

PR #423, opened 2021-05-31
bug_422: Fix sobol conf_level range check

scottshambaugh:
Fix for https://github.com/SALib/SALib/issues/422.


ConnectedSystems:
Hello, thanks for this. 

I'd like to add some tests to catch this in case of regression before accepting your PR.

I'm very pressed for time at the moment but promise to have a look soon.

PR #424, opened 2021-06-05
v1.4 release candidate merge

ConnectedSystems:


ConnectedSystems:
Note that this version update includes #414 which requires that the number of samples be base 2 for the `saltelli` method.

A `ValueError` is raised if this requirement is not met, but the check may be bypassed by setting `check_conv` to `False`.

I am of two minds on this. On one hand it ensures correct behaviour as outlined in the PR and accompanying issue.
On the other hand, it may be "surprising" behavior for users and "break" a lot of existing code, even if the fix is simply specifying the setting above. Perhaps it's better to relax the constraint and default `check_conv` to `False`.

Any thoughts?

jdherman:
Thanks @ConnectedSystems, this is an incredible effort that ties up a lot of loose ends. All looks great.

About the #414 issue, I'd be in favor of raising a warning rather than an error. As far as I understand it's not strictly wrong to use a non-base 2 sample size, but it could be suboptimal. As you pointed out, an error would break a lot of existing code. With a warning we wouldn't need the boolean flag either.

ConnectedSystems:
I've removed the strict convergence check in favour of a warning to preserve backwards compatibility.

Some low hanging fruit left for v1.4.1 and beyond:
- Parallel estimation of sensitivity indices for each quantity of interest (each Y value)
- Clearing a ton of warnings from recent versions of numpy (something like 90+ if I remember right)

PR #446, opened 2021-07-08
Sample with power of two to avoid warning

schmitts:


ConnectedSystems (approved):
LGTM! Thanks again @schmitts for all your input on the documentation - it really is very valuable

PR #457, opened 2021-08-07
Parallel analysis

### LINKED ISSUE #456: Add parallel analysis
Users are currently expected to apply the desired analysis method to every column in the result array.

```python
# example taken from readme
sobol_indices = [sobol.analyze(problem, Y) for Y in y.T]
```

The hybrid OO/functional approach introduced in v1.4 handles this [automatically](https://github.com/SALib/SALib/blob/4012c1313ee360d0cc3be0fba084f4f86048e2ea/src/SALib/util/problem.py#L302) but does so sequentially as in the above example.

The approach could be extended to allow parallel analyses.

ConnectedSystems:
Adds an `analyze_parallel` method, and cleans up REPL representation.

This PR is currently in progress, tests to ensure results are identical to the sequential approach are being finalized.

Closes #456 

UPDATE: 

This PR incorporates several inter-related changes which I did a poor job of keeping separate. Given the related nature of the changes however, I think it is acceptable.

1. The base `evaluate` and `analyze` methods now support the `nprocs` keyword argument. If provided, the method simply redirects the call to the `evaluate_parallel` and `analyze_parallel` methods. The motivation for maintaining the separation is primarily to reduce code clutter.
2.  Use of processors is capped to the available/detected cores (for model evaluation) and number of outputs (for analysis). i.e., no benefit if there is only a single `Y`.
3. Initial move towards the now recommended `default_rng()` function to produce method/context specific random seeds. This contrasts the earlier usage of `np.random.seed()` which sets global state which can cause issues. This is only completed for the `sobol.analyze` function.
4. Expanded examples/documentation where appropriate
5. Example `lake_problem` adjusted to accept a seed value

As with most parallel solutions, the benefit only becomes obvious when applied to sufficiently long running models or large datasets.

Example:

```python
from SALib.test_functions import lake_problem
from SALib import ProblemSpec


sp = ProblemSpec({
    'names': ['a', 'q', 'b', 'mean', 'stdev', 'delta', 'alpha'],
    'bounds': [[0.0, 0.1], [2.0, 4.5], [0.1, 0.45], [0.01, 0.05], [0.001, 0.005], [0.93, 0.99], [0.2, 0.5]],
    'outputs': ['max_P', 'Utility', 'Inertia', 'Reliability']
})

(sp.sample_saltelli(2**15)
    .evaluate(lake_problem.evaluate, nprocs=2)
    .analyze_sobol(nprocs=2, seed=101))

print(sp)
```

@jdherman @willu47 either of you keen to do a code review, or happy for this to be merged into v1.4.1 directly?

### DIFF
--- a/examples/Problem/multi_output.py
+++ b/examples/Problem/multi_output.py
@@ -18,7 +18,7 @@ sp = ProblemSpec({
     'outputs': ['max_P', 'Utility', 'Inertia', 'Reliability']
 })
 
-(sp.sample_saltelli(1000)
+(sp.sample_saltelli(1024)
     .evaluate(lake_problem.evaluate)
     .analyze_sobol())
 
--- a/examples/Problem/problem_spec.py
+++ b/examples/Problem/problem_spec.py
@@ -38,16 +38,25 @@ if __name__ == '__main__':
         .analyze_sobol(calc_second_order=True, conf_level=0.95))
     print("Time taken with 1 core:", time.perf_counter() - start, '\n')
 
+    # Same above example, but passing in specific functions
+    # (sp.sample(saltelli.sample, 25000, calc_second_order=True)
+    #     .evaluate(Ishigami.evaluate)
+    #     .analyze(sobol.analyze, calc_second_order=True, conf_level=0.95))
+
     # Samples, model results and analyses can be extracted:
     # print(sp.samples)
     # print(sp.results)
     # print(sp.analysis)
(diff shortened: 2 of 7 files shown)


jdherman:
@ConnectedSystems this looks great to merge, thanks. Maybe two warnings could be added: 1. if the user requests more processors than are available (for both analyze/evaluate), and 2. if parallelism is not used because there is only one output (analyze only).

PR #489, opened 2021-11-06
Expanded Morris docs

ConnectedSystems:
Expanded documentation for the Morris method (both sampling and analysis) as part of our drive to have more informative/beginner-friendly documentation.


willu47 (approved):
The docstrings add a very useful commentary to users on how to use and implement the Morris method. Thanks for the contribution!

PR #490, opened 2021-11-09
Expanded docs for the PAWN method

ConnectedSystems:


willu47 (approved):
Adds a useful explanatory docstring for the PAWN method which will aid users in selecting the method.

PR #496, opened 2022-01-20
Update hdmr.py

zjzh:
refactoring code with chained comparison which is more pythonic, concise, readable and efficient; how do think this change which has practical value？

### DIFF
--- a/src/SALib/analyze/hdmr.py
+++ b/src/SALib/analyze/hdmr.py
@@ -496,7 +496,7 @@ def f_test(Y, f0, Y_em, R, alpha, m1, m2, m3, n1, n2, n3, n):
         # Number of parameters of proposed model (order dependent)
         if i <= n1:
             p1 = m1        # 1st order
-        elif i > n1 and i <= (n1 + n2):
+        elif n1 < i <= n1 + n2:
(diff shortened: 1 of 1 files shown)


src/SALib/analyze/hdmr.py
```diff
@@ -496,7 +496,7 @@ def f_test(Y, f0, Y_em, R, alpha, m1, m2, m3, n1, n2, n3, n):
         # Number of parameters of proposed model (order dependent)
         if i <= n1:
             p1 = m1        # 1st order
-        elif i > n1 and i <= (n1 + n2):
+        elif n1 < i <= n1 + n2:
```
ConnectedSystems, line 499: I would prefer the brackets around n1 + n2 be kept, as this is clearer to me.

```python
elif n1 < i <= (n1 + n2):
```

ConnectedSystems:
Hi @zjzh 

I have no objection to including this if you make the requested change.

Thank you

src/SALib/analyze/hdmr.py
zjzh: Okay, I just added the  brackets. Thanks for your comments.

Issue #344, opened 2020-09-10
Non-zero variance with constant array affecting results (DMIM)


ConnectedSystems:
Just got burnt by this.

Suppose we have an output that is completely insensitive, e.g.

```
Y = [192.53, 192.53, 192.53, 192.53, 192.53, ...]
```

The `delta` method (DMIM) as implemented in SALib reports some sensitivity:

![image](https://user-images.githubusercontent.com/4075136/92689489-6f7d4480-f382-11ea-85fd-41a482809678.png)

(the x-axis is the number of samples)

Why, you ask?

In calculating the Sobol' indices, DMIM divides by the variance of `Y`.
But if `Y` is unchanging, then the variance should be `0.0` and it should error with a `ZeroDivision` exception, no?

Well, apparently not, due to precision/rounding error:

```
>>> np.var(Y)
8.077935669463161e-28
```

(there's an open discussion on this over at: https://github.com/numpy/numpy/issues/9631)

To resolve this, I've modified the relevant function (first 3 lines of function):

```
def sobol_first(Y, X, m):
    if (len(Y.shape) == 1) and np.allclose((Y - Y[0]), 0.0):
        # If Y does not change then it is not sensitive to anything...
        return 0.0

    xr = rankdata(X, method='ordinal')
    Vi = 0
    N = len(Y)
    Y_mean = Y.mean()
    for j in range(len(m) - 1):
        ix = np.where((xr > m[j]) & (xr <= m[j + 1]))[0]
        nm = len(ix)
        Vi += (nm / N) * ((Y[ix].mean() - Y_mean)**2)

    return Vi / np.var(Y)
```

which resolves the issue for my specific case (single output vector), but likely could be made more general...

This issue might (likely to?) affect other methods as well, I haven't gone through and checked yet.

ConnectedSystems:
I've since generalized the above to:


```python
def sobol_first(Y, X, m):
    # pre-process to catch constant array
    # see: https://github.com/numpy/numpy/issues/9631
    if Y.ptp() == 0.0:
        # Catch constant results
        # If Y does not change then it is not sensitive to anything...
        return 0.0

    xr = rankdata(X, method='ordinal')
    Vi = 0
    N = len(Y)
    Y_mean = Y.mean()
    for j in range(len(m) - 1):
        ix = np.where((xr > m[j]) & (xr <= m[j + 1]))[0]
        nm = len(ix)
[...]

PR #505, opened 2022-04-01
Sobol': Handle zero variance edge case

ConnectedSystems:
As in title - handle cases where there is no variance (e.g., a constant array) but the reported variance is > 0 due to computational error.


willu47:
Interesting case. Will you provide a test? This might be an issue with other methods as well.

ConnectedSystems:
Yes, will write a test for this.

Details are similar if not identical to this one:
https://github.com/SALib/SALib/issues/344

ConnectedSystems:
Added tests, but open for additional comments.

Tests could probably extended to cover all other methods in the future.

ConnectedSystems:
I believe this should be ready for review now.

tupui (approved):
Changes are straightforward and seem fine.

An extra thing to consider could be to have a warning when that happens. That could help some users diagnostic an issue with their modeling.

ConnectedSystems:
Added a warning: 

    Constant values encountered, indicating the model evaluations (or subset of evaluations) produced identical values.

The warning is raised when this condition is encountered for first and second order calculations. No warnings are raised when bootstrapping.

Would appreciate feedback on whether this message is informative enough.

PR #509, opened 2022-05-10
Better Triangular Distribution

BrandonSLockey:
Hi, Please consider this improved triangular distribution syntax for Sobol Sensitivity Analysis, previously user was forced to start at 0 when specifying triangular distribution. Now we can start at any number >0.

Need to Update the Docs: First Parameter is now Triangular Distribution Start, Second Parameter is Triangular Distribution End, 3rd parameter is position of Peak.

image

### DIFF
--- a/README-advanced.md
+++ b/README-advanced.md
@@ -124,11 +124,13 @@ parameter bounds but sample-specific metadata.
 
 * unif: uniform distribution
     e.g. :code:`[-np.pi, np.pi]` defines the lower and upper bounds
-* triang: triangular with width (scale) and location of peak. 
-    Location of peak is in percentage of width.
-    Lower bound assumed to be zero.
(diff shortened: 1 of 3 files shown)


ConnectedSystems:
Hi @BrandonSLockey 

Thanks for your PR. 
Would you happen to have, or be willing to make, a simple example to include as a test?

In the meantime, I will adjust things to still use the older syntax in the two-value case for backward compatibility. We can drop support for the older approach when we hit v2.0

ConnectedSystems:
After some thinking, I've decided this is probably stable enough to merge. Thanks for your contribution @BrandonSLockey !

BrandonSLockey:
> Thanks for your PR. Would you happen to have, or be willing to make, a simple example to include as a test?

sorry I didnt have time to make one, but thank you for implementing it 🙂

PR #511, opened 2022-05-15
HDMR Performance improvements

ConnectedSystems:
Principle change included in this PR is the replacement of `np.matmul` with `np.einsum` to avoid the repeated use of the former within loops, reducing runtime by ~35% compared to the original implementation with default values applied to the Sobol G function.

Other minor changes have also been made for brevity/readability, primarily by removing direct explicit use of `np.multiply`/`np.subtract`/etc for their implicit counterparts (e.g., `-`, `*` etc)


ConnectedSystems:
@sahin-abdullah please hold off doing a complete review at this stage, as I'm still working on providing better support for the ProblemSpec interface. Other comments welcome, however.

ConnectedSystems:
Hey @sahin-abdullah 

Tried the matrix multiplication outside the loops you did and saw the same performance regression you did.
I don't think it's a memory copy issue like I suspected earlier. Might have to do with numpy not using the most optimal library (e.g., BLAS vs MKL) for this case. I'm not too familiar with numpy internals so can't comment further.

I've taken the simplest path and just reverted to the version that does the multiplications within the loops for now.

image

ConnectedSystems:
Thanks @sahin-abdullah !

PR #514, opened 2022-07-11
Nonuniform distributions for Morris sampling returns inf

jyangfsu:
When the parameters follows norm, lognorm and truncnorm distributions, nonuniform_scale_samples function would returns inf. 
As suggested by Saltelli et al. (2010), this can be avoid by cutting the tails of , for example, the normal distributions, at quantiles 5 and 95%.
Code in current nonuniform_scale_samples function:
elif dists[i] == 'norm':
            if b2 <= 0:
                raise ValueError("""Normal distribution: stdev must be > 0""")
            else:
                conv_params[:, i] = sp.stats.norm.ppf(
                    params[:, i], loc=b1, scale=b2)

Suggest modifying to:
        elif dists[i] == 'norm':
            if b2 <= 0:
                raise ValueError('''Normal distribution: stdev must be > 0''')
            else:
                conv_params[:, i] = scipy.stats.truncnorm.ppf(
                    params[:, i], scipy.stats.norm.ppf(0.05,  \
                          loc=b1, scale=b2), scipy.stats.norm.ppf(0.95,  \
                          loc=b1, scale=b2), loc=b1, scale=b2)


willu47:
Converted to issue #515

Issue #518, opened 2022-08-05
Question: Can I add samples to an existing Saltelli sequence?


MarAlder:
Hi,

I really enjoy applying SALib for Sensitivity Analyses. From a user's point of view, I have the following question:

### Problem description
Let's say I have the following problem specification:
```python
problem = {
        "num_vars": 2,
        "names": ["var1", "var2"],
        "bounds": [[-2000.0, 2000.0], [0.90, 1.1]],
    }
```
I generate a LDS via:
```python
saltelli.sample(problem, N, calc_second_order=False)
```
to compute the Sobol indices:
```python
sobol.analyze(problem, Y, calc_second_order=False)
```

### Question
If run the analysis to compute `Y` and then conclude that I would need, e.g., `N=1024` samples rather than `N=512` to get a satisfactory convergence of my Sobol indices, can I re-use the previous samples and calculations by extending the existing sequence? I'm not too deep into the theory of LDS, but found something like this at [qmc](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.qmc.Sobol.html#scipy.stats.qmc.Sobol) for Sobol sequences (see `To continue an existing design, extra points can be obtained by [...]`). 

I played with the sampler and found that the first 16 values generated by `saltelli.sample(problem, 4, calc_second_order=False)` and `saltelli.sample(problem, 16, calc_second_order=False)` are equal. However, the first 64 values of `saltelli.sample(problem, 16, calc_second_order=False)` and `saltelli.sample(problem, 32, calc_second_order=False)` do not match. Therefore, I wonder if I must compute all samples again?

I appreciate your response! Thanks,
Marko

PS: I think I don't have the rights to add the `question_interpretation` label.

tupui:
Hi @MarAlder this is because of the `skip_values` parameter which is "kicking in". Otherwise samples from the small sample should be present in the other one. Otherwise Sobol' being a sequence, you are able to reuse the previous samples as Saltelli's method is just a shuffling on the dimensions, not the number of samples.

MarAlder:
Ah ok, thank you!

PR #519, opened 2022-08-05
Add Sobol' sampler from SciPy

### LINKED ISSUE #450: Implement scrambling of Sobol' sequence
Scrambled Sobol' sequences offer improved uniformity and convergence properties compared with earlier approaches.

http://jcgt.org/published/0009/04/01/

http://statweb.stanford.edu/~owen/reports/dropzero.pdf

### LINKED ISSUE #478: Add new column shift method for Sobol sampling
Recently published column shift method by @Xifus could be incorporated into SALib

Repo can be found here: https://github.com/Xifus/ColumnShiftMethod

### LINKED ISSUE #486: Proposal: Adopt scipy Sobol and deprecate `sample.saltelli`
Scipy now offers a [Sobol' sampling method](https://scipy.github.io/devdocs/reference/generated/scipy.stats.qmc.Sobol.html) which includes scrambling and supports a more computationally efficient approach to continuing an existing design (skipping values).

I propose we wrap the scipy method which would effectively resolve #450 and #479.

Leveraging the scipy implementation would allow us to focus on additional Sobol approaches not currently available elsewhere, such as #478.

Implications:
- The sobol analysis method may need adjustment.
- The additional functionality offered by the scipy method (e.g., discrepancy criterion, sampler resetting, and others) would not be directly accessible for those using the procedural approach. There may be ways to mitigate this, however.
- Additional infrastructure would need to be designed for the object-oriented interface, although this may be relatively minor 
- If this proposal is adopted, the current `sample.saltelli` method would be deprecated rather than modified to maintain backward compatibility. I suggest removing the saltelli implementation in v2.0

Additional dependencies: 
None, as Scipy is already a dependency

pinging @tupui @Xifus @sahin-abdullah in case they have thoughts

tupui:
Closes #486, closes #450 and should close #478 as if the method does work better (which I am not convinced yet from my tests) should go in SciPy directly.

Following discussions in #486, this adds Sobol' sampler as a new sampler and deprecate `salib.sample.saltelli`.

Note that SciPy also has LHS (and Halton that can be used otherwise).

This needs SciPy to 1.7 and doing so I updated NumPy and Python minimal versions following NEP29.

This is a draft as I need some guidance on how to proceed next. E.g. I removed the skipping as we can scramble and should have better properties doing so.

cc @ConnectedSystems

### DIFF
--- a/src/SALib/sample/sobol.py
+++ b/src/SALib/sample/sobol.py
@@ -5,9 +5,7 @@ import numpy as np
 from scipy.stats import qmc
 
 from . import common_args
-from ..util import (
-    scale_samples, read_param_file, compute_groups_matrix, _check_groups
-)
+from ..util import scale_samples, read_param_file, compute_groups_matrix, _check_groups
 
 
 def sample(
@@ -17,7 +15,7 @@ def sample(
     calc_second_order: bool = True,
     scramble: bool = True,
     skip_values: int = 0,
-    seed: Optional[Union[int, np.random.Generator]] = None
+    seed: Optional[Union[int, np.random.Generator]] = None,
 ):
     """
     Generates model inputs using Saltelli's extension of the Sobol' sequence.
@@ -98,16 +96,16 @@ def sample(
            The Second IMACS Seminar on Monte Carlo Methods 55, 271–280.
            https://doi.org/10.1016/S0378-4754(00)00270-6
     """
-    D = problem['num_vars']
+    D = problem["num_vars"]
     groups = _check_groups(problem)
 
     # Create base sequence - could be any type of sampling
-    qrng = qmc.Sobol(d=2*D, scramble=scramble, seed=seed)
+    qrng = qmc.Sobol(d=2 * D, scramble=scramble, seed=seed)
 
     # fast-forward logic
     if skip_values > 0:
         M = skip_values
-        if not ((M & (M-1) == 0) and (M != 0 and M-1 != 0)):
+        if not ((M & (M - 1) == 0) and (M != 0 and M - 1 != 0)):
             msg = f"""
             Convergence properties of the Sobol' sequence is only valid if
             `skip_values` ({M}) is a power of 2.
@@ -134,7 +132,7 @@ def sample(
     base_sequence = qrng.random(N)
 
     if not groups:
-        Dg = problem['num_vars']
+        Dg = problem["num_vars"]
     else:
         G, group_names = compute_groups_matrix(groups)
         Dg = len(set(group_names))
@@ -156,8 +154,7 @@ def sample(
         # Cross-sample elements of "B" into "A"
         for k in range(Dg):
             for j in range(D):
-                if (not groups and j == k) or \
-                   (groups and group_names[k] == groups[j]):
+                if (not groups and j == k) or (groups and group_names[k] == groups[j]):
                     saltelli_sequence[index, j] = base_sequence[i, j + D]
                 else:
                     saltelli_sequence[index, j] = base_sequence[i, j]
@@ -169,8 +166,9 @@ def sample(
         if calc_second_order:
             for k in range(Dg):
                 for j in range(D):
-                    if (not groups and j == k) or \
-                       (groups and group_names[k] == groups[j]):
+                    if (not groups and j == k) or (
+                        groups and group_names[k] == groups[j]
+                    ):
                         saltelli_sequence[index, j] = base_sequence[i, j]
                     else:
                         saltelli_sequence[index, j] = base_sequence[i, j + D]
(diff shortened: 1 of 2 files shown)


tupui:
I tried to update the ubuntu version in the CI but it would need more work. So probably better not to touch this here.

ConnectedSystems:
Thanks @tupui 

I will take a closer look when I can (may take a week or two) but at first glance this looks like a very clean re-implementation.

Re skipping vs scrambling, I would prefer to support skipping anyway, as I think the user should be able to decide what to apply. 

One thing I would like to see is expanded docstrings with more complete notes to cover the differences (I can do this if you like).

tupui:
Great thanks @ConnectedSystems. It took me quite some time to do this as well 😉

> I will take a closer look when I can (may take a week or two) but at first glance this looks like a very clean re-implementation.

It's highly based on the current Saltelli sampler. I just removed the skipping part.

> Re skipping vs scrambling, I would prefer to support skipping anyway, as I think the user should be able to decide what to apply.

But what is the use case? As seen in #518. Skipping bring difficulties if one want's to resample. And it's also a complex logic to keep the properties.

Instead, if the use case is to have different samples, I would suggest adding a `seed` or `random_state` parameter. This way users can increment this instead and the scrambler does the correct thing behind the scene.

This brings up the `scramble` parameter. I added this almost just as a provision in case you want to set a different strategy, but as it is, it should always be True. There is no operational reasons to not scramble the sequence.

> One thing I would like to see is expanded docstrings with more complete notes to cover the differences (I can do this if you like).

Sure, I can do this once we settle on the API.

ConnectedSystems:
> But what is the use case? As seen in https://github.com/SALib/SALib/issues/518. Skipping bring difficulties if one want's to resample. And it's also a complex logic to keep the properties.

I understand the reservation - my view is that SALib is useful as a teaching tool as well as a research tool. Someone out there may be relying on it for some purpose (e.g., to show case different strategies). The code itself is useful if one were interested in seeing how skipping might be implemented.

> Instead, if the use case is to have different samples, I would suggest adding a `seed` or `random_state` parameter

I think `seed` is best as it matches implementation of other methods.

tupui:
Ok I can add back the skip parameter following what's in the current Saltelli method. And I will add seed.

src/SALib/sample/sobol.py
```diff
@@ -0,0 +1,234 @@
...
+           https://doi.org/10.1016/S0378-4754(00)00270-6
+    """
+    D = problem['num_vars']
+    groups = _check_groups(problem)
+
+    # Create base sequence - could be any type of sampling
+    qrng = qmc.Sobol(d=2*D, scramble=scramble, seed=seed)
+
+    # fast-forward logic
+    if skip_values > 0:
```
ConnectedSystems, line 107: Is it acceptable if `skip_values > 0` and `scramble == True`?

src/SALib/sample/sobol.py
```diff
@@ -0,0 +1,234 @@
...
+        Dg = problem['num_vars']
+    else:
+        G, group_names = compute_groups_matrix(groups)
+        Dg = len(set(group_names))
+
+    if calc_second_order:
+        saltelli_sequence = np.zeros([(2 * Dg + 2) * N, D])
+    else:
+        saltelli_sequence = np.zeros([(Dg + 2) * N, D])
+    index = 0
```
ConnectedSystems, line 145: Please add line between `if` block and `index = 0` for clarity

src/SALib/sample/sobol.py
```diff
@@ -0,0 +1,234 @@
...
+
+    for i in range(N):
```
ConnectedSystems, line 147: Not completely necessary but I suggest we replace these nested loops with `itertools.product()`

src/SALib/sample/sobol.py
ConnectedSystems: Whatever the answer, it would be worthwhile to add it to the docstring I think

ConnectedSystems (commented):
Thanks, just some minor changes and a question for you.

src/SALib/sample/sobol.py
tupui: Done

src/SALib/sample/sobol.py
tupui: Yes we can combine both. I added a note in the docstrings as suggested.

src/SALib/sample/sobol.py
tupui: I am not sure how to go about this. The loops need to update `index` in the first level. Maybe this can be kept for latter. And actually there might be another way to write these with NumPy operations.

tupui (commented):
Thanks for the review @ConnectedSystems. I updated the PR

tupui:
@ConnectedSystems the CI failure is also present on `main`. I propose to fix this in another PR as I update the CI to use GitHub Actions.

tupui:
I updated the PR with the recent changes in `main`. Dependencies are already up to date now.

PR #522, opened 2022-08-17
Make Pathos optional

### LINKED ISSUE #521: Make Pathos an optional dependency
As far as I can see in the code, Pathos is only required in a single function to run distributed runs.

I propose to make it an optional dependency to lower the footprint of the library.

It should just consist in a condition at the top checking for the import, set a flag and then in the function if would use the flag to report a missing deps.

I can do the work if wanted.

tupui:
Closes #521

This removes the dependency to Pathos to make it optional.

We could also do something similar for `multiprocess` (which is pulled by `Pathos` at the moment and used for parallel execution.). In that case, it should be fairly easy to write compatible code that would use `multiprocess` if installed and fallback to `multiprocessing` if not.

I am not sure about how the packaging is handled, so I did not update for now the requirements files (and I would suggest moving to something simpler like `flit` or `hatch` which is now the recommendation for pure Python packages (see https://packaging.python.org/en/latest/tutorials/packaging-projects/), I can do the lifting here if wanted).

### DIFF
--- a/examples/Problem/problem_spec.py
+++ b/examples/Problem/problem_spec.py
@@ -78,6 +78,9 @@ if __name__ == "__main__":
     print(sp)
 
     # Distributed example
+    # This example requires an additional dependency (Pathos) to be installed
+    # `pip install pathos` or `conda install pathos -c conda-forge`
+
     # Specify itself as servers as an example
     servers = ("localhost:55774", "localhost:55775", "localhost:55776")
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,11 +35,16 @@ dependencies = [
   "scipy>=1.7.3",
   "matplotlib>=3.2.2",
   "pandas>=1.1.2",
-  "pathos>=0.2.5",
+  "multiprocess",
 ]
 
 [project.optional-dependencies]
+distributed = [
+    "pathos>=0.2.5",
+]
+
 test = [
+    "SALib[distributed]",
     "pytest",
     "pytest-cov",
 ]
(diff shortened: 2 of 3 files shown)


tupui:
I went forward and also made `multiprocess` optional since it's a one liner change.

tupui:
Ok this fails because of this

```python
with Pool(nprocs) as pool:
    res = list(pool.imap(lambda y: func(self, Y=y),
               [self._results[:, i] for i in range(Yn)]))
```

lambda functions cannot be serialized and this is where multiprocess is helping. This being just 1 test. I am not sure how this code path is critical/optional. Let me know how you want to proceed.

ConnectedSystems:
Thanks for jogging my memory @tupui 

I think leave `multiprocess` in as a core dependency as it provides support for parallel model evaluation and analysis "out of the box". I don't recall it being a heavy package either.

Pathos can remain as an optional dependency given its niche use case.

ConnectedSystems (approved):
All looks good to me, I just added a note in the example file for distributed evaluations.

Thanks!

ConnectedSystems:
Tests fail but it occurs while collecting and installing packages. Not sure what the issue is but it's too late for me to dig further. I'll have another look when I can.

ConnectedSystems:
Seems TravisCI fails as Python 3.7 has been removed

image

I will delay merging this until #519 is ready as it looks like you've already made the necessary updates to the env config there.

tupui:
I updated the PR with the recent changes in `main`.

PR #523, opened 2022-08-18
Use pydata-sphinx-theme

tupui:
Closes #520, closes #531

cc @sahin-abdullah

PR #527, opened 2022-08-24
Migrate to hatchling and migrate CI to GitHub actions

### LINKED ISSUE #355: Document release process
1. Write down list of steps to take when preparing for a new release
2. Create an Issue Template which can be used to keep track of these steps for each release (e.g. a new issue using the template is opened when preparing a release, and closed upon release).

### LINKED ISSUE #524: Consider/investigate alternate packaging tool(s)
Raised in #522 , there are newer packaging tools (Flit, Hatchling) that may help reduce maintenance overhead

https://packaging.python.org/en/latest/tutorials/packaging-projects/

cc @willu47 @tupui 

Primary question that comes to mind is: are there any pros/cons with our current setup vs these newer tools (e.g., if we end up incorporating Cython)

tupui:
Closes #524 and closes #355

This does 2 things:

* Use Hatchling and doing so removes all extra files which are not needed anymore.
* Migrate from Travis to GitHub Actions: 3 actions are used for linting, testing and deploying.

#519 and #522 would need to be slightly adapted depending on which PR gets in first.

Also, PyPi secrets would need to be added to the repo for deploying.

I simplified the version handling as well. I don't think there is a need to complicate this more. Also note that this is more or less for legacy use as the new way to check that is to rely on package's metadata.


tupui:
I could fix the linter in one commit if wanted.

tupui:
@ofek could you have a quick look at the Hatchling config? This project is using a src layout and for some reason I cannot manage to properly import the package. If I remove the src folder then all works just fine. I tried to setup a blank project with the layout configuration option set to True and it was working fine. So I am not sure what is wrong here. Can it be due to the casing? Something else I am missing?

pyproject.toml
```diff
@@ -1,8 +1,81 @@
...
+[project.scripts]
+salib = "SALib.scripts.salib:main"
+
+[project.urls]
+Documentation = "https://salib.readthedocs.io"
+"Source code" = "https://github.com/SALib/SALib"
+
+[tool.hatch]
+version.path = "src/SALib/__init__.py"
+build.exclude = [
```
ofek, line 65:
```suggestion
build.targets.sdist.exclude = [
```

tupui:
Oh I got it to work doing `build.targets.wheel.packages = ["src/SALib"]`. I did not get that from the doc.

pyproject.toml
tupui: Ah ok I thought I could do this as seen here so that it would be done for both sdist and wheel https://hatch.pypa.io/latest/config/build/#patterns

ofek:
I just tried building from before those 2 commits and seems fine

w1

w2

tupui:
I did not try to build before that, I was just installing with `pip install -e .` and then trying to do `import SALib` and this was failing. It's only with the last commit that the import was working. And now we see the tests are passing.

pyproject.toml
ofek: Oh yeah you can I just thought you meant sdist

ofek:
Oh wow you discovered a bug: `salib` vs `SALib`

I'll fix tonight

tupui:
@willu47 @ConnectedSystems this is ready to be reviewed. The commit [`0347916` (#527)](https://github.com/SALib/SALib/pull/527/commits/03479161e6970c66a9b02a7b24e88bae27674f04) is running the commit hooks hence the large diff. I suggest not looking at this commit when reviewing (in case you don't know, on GitHub you can filter commit's to review. Under conversation you have a `all commits` drop down). If you prefer I can remove this commit for now.


Screenshot 2022-08-24 at 19 49 11

ofek:
Please remove the explicit `build.targets.wheel.packages` and try https://github.com/pypa/hatch/releases/tag/hatchling-v1.8.1

tupui:
Thank you @ofek for the fast update of Hatchling! It works just fine now. I tested locally and in the CI here. Thank you for having a look at this configuration and helping out.

ofek:
Happy to help!

pyproject.toml
```diff
@@ -1,8 +1,86 @@
...
+requires = ["hatchling"]
+build-backend = "hatchling.build"
 
-[tool.setuptools_scm]
-# See configuration details in https://github.com/pypa/setuptools_scm
-version_scheme = "no-guess-dev"
+[project]
+name = "SALib"
+dynamic = ["version"]
+description = "Tools for sensitivity analysis. Contains Sobol', Morris, and FAST methods"
```
willu47, line 8: Update the description with the complete list of methods provided. Add 'global' in front of 'sensitivity analysis'

pyproject.toml
```diff
@@ -1,8 +1,86 @@
...
+readme = "README.rst"
+requires-python = ">=3.8"
+license = "MIT"
+authors = [
+  { name = "Jon Herman", email = "jdherman8@gmail.com" },
+  { name = "Will Usher" },
```
willu47, line 14: Please add @ConnectedSystems to the author list.

pyproject.toml
```diff
@@ -1,8 +1,86 @@
...
+]
+maintainers = [
+  { name = "SALib contributors" },
```
willu47, line 17: Should we be more specific here and take the list from github?

willu47 (changes requested):
Thanks for updating the CI and CD steps.

Please could you ensure that the developer documentation in `CONTRIBUTING.md` is also updated to reflect the new CI/CD procedures? (note @ConnectedSystems should be listed as an admin there as well). 

It would also be useful to record the steps required to create a release (I believe I wrote this up in an issue or PR a long time ago!).

pyproject.toml
tupui: I would not really advise for it as it's a lot of additional maintenance. I think there are bots that could help do things like these. But usually projects care more about updating the release notes and have all the folks there. I suggest doing this elsewhere in any case as this PR is already packed with a lot of things.

tupui:
Thank you for the review @willu47!

I did the updates. For the documentation, I did already update the install instructions and added a section about creating an env with SALib installed in dev mode. 

For the release process (then this should close #355), I added the steps at the end of the contributing doc.

In general, I do think that the general structure and text could be updated. Though, I would suggest to separate theses changes and have them over #523 where I started to update the theme (which is ready on my side and needs inputs).

ofek (approved):

ConnectedSystems:
Hi @tupui 

Tried to do a local install and test run and found these two issues:

1. Instructions on how to run tests need to updated. On that, should there be a script that supports `hatch run tests` or is simply `pytest` okay?
2. When running tests, I get a circular import error. I haven't dug deeply on it at all but my first guess is that this is due to new behavior that relies on SciPy in modules that did not previously require it. 

An example of one of the test errors:

```bash
Traceback:
..\..\..\miniconda3\envs\salib\lib\importlib\__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests\test_sample_seed.py:5: in <module>
    from SALib.sample.morris import sample as morris_sampler
src\SALib\sample\morris\__init__.py:1: in <module>
    from .local import LocalOptimisation  # noqa: F401
src\SALib\sample\morris\local.py:7: in <module>
    from .strategy import Strategy
src\SALib\sample\morris\strategy.py:16: in <module>
    from scipy.spatial.distance import cdist  # type: ignore

[... snip ...]

ImportError: cannot import name '_iterative' from partially initialized module 'scipy.sparse.linalg.isolve'
(most likely due to a circular import) 
(C:\[path to SALib environment]\lib\site-packages\scipy\sparse\linalg\isolve\__init__.py)
```

tupui:
Interesting I don't recall having issues there, and the CI is passing just fine. I would suspect an issue with your env and maybe try from scratch following what is done in the CI. I will try to reproduce on Monday.

Since we are using conda, there is no reason to me to use `hatch run`. My understanding of the command is that it would do something similar, ensuring you are using an env which matches your config. So yes just running `pytest` while being in a conda env is enough.

ConnectedSystems:
Okay, this is a weird one.

I run into an ImportError with SciPy due to a circular dependency

```
In [1]: import scipy
In [2]: scipy.stats

---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Input In [6], in <cell line: 1>()
----> 1 scipy.stats

File C:\programs\miniconda3\envs\salib\lib\site-packages\scipy\__init__.py:211, in __getattr__(name)
    209 def __getattr__(name):
    210     if name in submodules:
--> 211         return _importlib.import_module(f'scipy.{name}')
    212     else:
    213         try:

# ... snip ...

ImportError: cannot import name '_iterative' from partially initialized module 'scipy.sparse.linalg.isolve' (most likely due to a circular import) (C:\[salib environment location]\lib\site-packages\scipy\sparse\linalg\isolve\__init__.py)
```

This is a fresh environment but will retry with another one. And maybe again on Linux, given the tests pass.

tupui:
Sounds like your env is corrupted. Although from the traceback I am not sure if `C:\[salib environment location]` matches `C:\programs\miniconda3\envs\salib`

ConnectedSystems:
False alarm - all working now.

Had to recreate the environment again after manually deleting the environment folder but its working now.

ConnectedSystems:
All looks good to me so am merging now.

Thanks to all who took part, and again to @ofek for dropping by and helping out!

Issue #532, opened 2022-09-08
Large number of Sphinx warnings/errors when building documentation


willu47:
Okay, I reinstalled my conda environment, built the docs locally and had a brief scan through.  The new theme looks great, and nice that we're joining the look-and-feel of the pydata stack.

```bash
conda env create -f environment.yml
conda activate salib
cd docs
make html
open _build/html/index.html
```
Sphinx itself spits out a lot of warnings and a few errors which we should slowly look into:

```python
Running Sphinx v5.1.1
Creating file docs/api/SALib.rst.
Creating file docs/api/SALib.analyze.rst.
Creating file docs/api/SALib.plotting.rst.
Creating file docs/api/SALib.sample.rst.
Creating file docs/api/SALib.sample.morris.rst.
Creating file docs/api/SALib.scripts.rst.
Creating file docs/api/SALib.test_functions.rst.
Creating file docs/api/SALib.util.rst.
Creating file docs/api/modules.rst.
loading pickled environment... failed
failed: No module named 'myst_parser.main'
loading intersphinx inventory from https://www.sphinx-doc.org/en/master/objects.inv...
loading intersphinx inventory from https://docs.python.org/3.10/objects.inv...
loading intersphinx inventory from https://matplotlib.org/stable/objects.inv...
loading intersphinx inventory from https://numpy.org/doc/stable/objects.inv...
loading intersphinx inventory from https://scikit-learn.org/stable/objects.inv...
loading intersphinx inventory from https://pandas.pydata.org/pandas-docs/stable/objects.inv...
loading intersphinx inventory from https://docs.scipy.org/doc/scipy/objects.inv...
loading intersphinx inventory from https://pyscaffold.org/en/stable/objects.inv...
[autosummary] generating autosummary for: advanced.rst, api.rst, api/SALib.analyze.rst, api/SALib.plotting.rst, api/SALib.rst, api/SALib.sample.morris.rst, api/SALib.sample.rst, api/SALib.scripts.rst, api/SALib.test_functions.rst, api/SALib.util.rst, api/modules.rst, authors.rst, basics.rst, changelog.md, citations.rst, getting-started.rst, index.rst, license.rst
[...]

PR #540, opened 2022-10-08
WIP: Documentation fixes for Sphinx and website

ConnectedSystems:
Work in progress PR.

Number of warnings/errors currently stand at 34 with these changes. 

Addresses #532


src/SALib/sample/sobol.py
```diff
@@ -31,35 +31,38 @@ def sample(
...
+
     These model inputs are intended to be used with
     :func:`SALib.analyze.sobol.analyze`.
 
+
     Notes
     -----
     The initial points of the Sobol' sequence has some repetition (see Table 2
-    in Campolongo [1]_), which can be avoided by scrambling the sequence.
+    in Campolongo [1]__), which can be avoided by scrambling the sequence.
```
tupui, line 42: This should have been ok with the way references where listed in the reference section.

The fix in general works, an alternative was to make use of the refs in the text. Still I am surprised it was actually an issue in the first place as we have that in SciPy and don't call all the refs in the text.

ConnectedSystems: I wonder what plugins are in use for SciPy for the number style referencing?

At the moment, we get a large number of errors related to duplicate or unused references because the plugin expects the references to be globally unique. Most of the suggested solutions seem to be to switch to using bibtex style referencing, or to remove the reference system/plugin and just use simple numbered lists.

tupui: @ConnectedSystems sorry for the delay. Right this is coming from numpydoc, see: https://numpydoc.readthedocs.io/en/latest/format.html#references

Maybe something to consider?

ConnectedSystems: Ah, so any references has to be used under the  `Notes` section? That can be done I guess...

tupui: It's not an obligation. You can have refs which are not being called (we have plenty in SciPy). Although might make sense long term to try to use what we cite, like in an paper.

Also you can call refs in the whole docstring, not only the Notes section.

willu47 (changes requested):
Hi @ConnectedSystems - please use [semver](http://semver.org) for version numbers!

ConnectedSystems:
Hi @willu47 

Yes, sorry - seems my understanding of semver is either outdated or I was misinformed from the beginning. I will endeavor to follow the official semver spec from now.

willu47:
> Hi @willu47
> 
> Yes, sorry - seems my understanding of semver is either outdated or I was misinformed from the beginning. I will endeavor to follow the official semver spec from now.

I'm just nitpicking! It's not a big deal, but it does avoid issues further down the line. For example Pypi tends to be quite intolerant of version numbers that don't follow their conventions.  Semver works fine with PyPi, although pypi does require a particular format for alpha or beta versions (see [PEP440](https://peps.python.org/pep-0440/)).

PEP 440 does allow four levels `v1.2.3.4` but it's generally not worth it! Tweaks to documentation are fine to be contained within the 3rd level.

tupui:
To replace 1.4.6.1, you could yank it on PyPi and make a 1.4.7

ConnectedSystems:
Errors/warnings reduced down to 21, mostly to do with duplicate descriptions, unexpected indentations (which I can't seem to resolve) and "unused references".

I think these can be fixed later when I/someone has more energy to look into it.

Included in this PR are a number of updates to the tutorials - primarily so that usage of the SALib Interface is properly showcased such that the documentation now matches with the examples in the recent paper.

Given docs are fairly important, I will merge and release v1.4.7 shortly

PR #546, opened 2022-10-28
Fix log2 in Sobol' exponent check

### LINKED ISSUE #545: skip_values error in sobol.sample
I noticed the new sobol.sample has an error when using skip-values. The error comes from line 117 (and 118). 
np.log(N,2) raises an error, because log base 2 is done using np.log2(N). The second argument is not the base but rather an output array.

tupui:
Closes #545

Definitely need to change the tests to run this instead of the older Saltelli sampler.


ConnectedSystems:
Beat me to it, thanks @tupui

PR #551, opened 2023-01-10
Update flake8 repository

tupui:
Flake8 was moved from GitLab to GitHub. Hence the CI failing.


tupui:
Linter is now failing but because it should be run again on the repo. Not anymore because it's just not running.

PR #556, opened 2023-03-17
Better, and explicit, support for IPython console display

### LINKED ISSUE #555: ProblemSpec.sample() causes permanent message
I'm just starting to get into `SALib` and I'm drawn to the `ProblemSpec` class, which provides easy access to all the main features of the library. However, the `sample()` method causes a message to appear forever. 

I'm using the stable version (via `pip install SALib`), so this may not be an issue with the github version. 

Here's some reproducible code (interactive session), with output in comment blocks:
```python
import numpy as np
from SALib import ProblemSpec
from SALib.sample import sobol

prob = ProblemSpec({
        "names": ["x1", "x2", "x3"],
        "groups": None,
        "bounds": [[-np.pi, np.pi]] * 3,
        "outputs": ["Y"],
})

prob.sample(func=sobol.sample, N=1024)
# {'names': ['x1', 'x2', 'x3'], 'groups': None, 'bounds': [[-3.141592653589793, 3.141592653589793], [-3.141592653589793, 3.141592653589793], [-3.141592653589793, 3.141592653589793]], 'outputs': ['response'], 'num_vars': 3, 'sample_scaled': True}
# Samples:
# -- 3 parameters: ['x1', 'x2', 'x3']
# -- 8192 evaluations 

prob.samples
# array([[ 0.27913238, -2.16263456, -1.65803756],
#        [-1.82017621, -2.16263456, -1.65803756],
#        [ 0.27913238,  3.13334766, -1.65803756],
#        ...,
#        [-3.0968241 ,  2.8720456 ,  1.50480474],
#        [-3.0968241 ,  1.73726994,  1.42794031],
#        [-3.0968241 ,  1.73726994,  1.50480474]])
# Samples:
# -- 3 parameters: ['x1', 'x2', 'x3']
# -- 8192 evaluations 

print(7)  # completely unrelated
# 7
# Samples:
# -- 3 parameters: ['x1', 'x2', 'x3']
# -- 8192 evaluations 
```

Is this normal? I don't get the same issue if I run `sobol.sample(prob, 1024)` instead.

----

Looking through the source code, it appears that this caused by the `__str__` method, which prints instead of returning a string.

ConnectedSystems:
Closes #555

### DIFF
--- a/src/SALib/util/problem.py
+++ b/src/SALib/util/problem.py
@@ -594,6 +594,9 @@ class ProblemSpec(dict):
                 method_name, MethodType(self._method_creator(func, "analyze"), self)
             )
 
+    def _repr_pretty_(self, p, cycle):
+        p.text(str(self) if not cycle else "...")
+
     def __str__(self):
         rep = ""
         if self._samples is not None:
--- a/src/SALib/util/problem.py
+++ b/src/SALib/util/problem.py
@@ -595,44 +595,51 @@ class ProblemSpec(dict):
             )
 
     def __str__(self):
+        rep = ""
         if self._samples is not None:
             arr_shape = self._samples.shape
             if len(arr_shape) == 1:
                 arr_shape = (arr_shape[0], 1)
             nr, nx = arr_shape
-            print("Samples:")
-            print(f"\t{nx} parameters:", self["names"])
-            print(f"\t{nr} evaluations", "\n")
+            rep += (
+                "Samples:\n"
+                f"\t{nx} parameters: {self['names']}\n"
+                f"\t{nr} evaluations\n"
+            )
         if self._results is not None:
             arr_shape = self._results.shape
             if len(arr_shape) == 1:
                 arr_shape = (arr_shape[0], 1)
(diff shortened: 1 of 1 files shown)


tupui (approved):
Changes LGTM 👍 (I did confirm that I could reproduce the behaviour in PyCharm with a dummy class. Very interesting that this happened!)

I did not realised you had some IPython code, that's great! I did some experiment at some point with that for SciPy and having some plots, etc. Something to consider maybe since we have plot methods already? (https://github.com/tupui/scipy-magic)

ConnectedSystems:
Thanks for the review! And I agree, seems PyCharm console is acting up.

Re `IPython magic` : This is a neat idea (and you're very prolific!).

I was also thinking of a separate companion visualization package, something that provides a dashboard-like experience, similar to GeoPanda's `.explore()`:
- https://geopandas.org/en/stable/docs/user_guide/interactive_mapping.html
- https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.explore.html

PR #558, opened 2023-03-19
hdmr maxorder=1 fix

eyyarbasi:
**hdmr** method was throwing a `keyerror` for `c2` if `maxorder = 1`

`keyerror` and corresponding `referencing before assignment` fixed.


ConnectedSystems:
Thanks for looking into this, I have just one comment.

@sahin-abdullah feel free to take a look as well if you have time/inclination.

ConnectedSystems:
Thanks everyone!

PR #563, opened 2023-04-05
Add discrepancy analysis

tupui:
Add `SALib.analyze.discrepancy` which is the method described in https://arxiv.org/abs/2206.13470

Based on 2D sub projections of [Xi,Y], the discrepancy of each samples is calculated which gives a value for all Xi. This can be used as a sensitivity indice.

Opening this in draft to get early feedbacks on how to proceed next. It's mainly missing tests and doc, but the feature itself should already work.

cc @arnaldpuy

### DIFF
--- a/src/SALib/analyze/discrepancy.py
+++ b/src/SALib/analyze/discrepancy.py
@@ -39,6 +39,22 @@ def analyze(
     Compatible with:
         all samplers
 
+    Based on 2D sub projections of ``[Xi,Y]``, the discrepancy of each sample
+    is calculated which gives a value for all ``Xi``. This information is used
+    as a sensitivity indice.
+
+    This measure is very fast and is visually explainable. Considering two
+    variables ``X1`` and ``X2``, ``X1`` is more influential than ``X2`` when
+    the scatterplot of ``X1`` against ``Y`` displays a more discernible shape
+    than the scatterplot of ``X2`` against ``Y``.
+
+    For the method to work properly, the input parameter space need to be
+    uniformely covered as the quality of the measure depends on the value of
+    the discrepancy. Taking a 2D sub projection, if the distribution of sample
+    along ``Xi`` is not uniform, it will have an impact on the discrepancy,
+    the value will increase, i.e. the importance of this parameter would be
+    inflated.
+
     References
     ----------
     1. A. Puy, P.T. Roy and A. Saltelli. 2023. Discrepancy measures for
--- a/src/SALib/analyze/discrepancy.py
+++ b/src/SALib/analyze/discrepancy.py
@@ -21,18 +21,16 @@ def analyze(
     ----------
     problem : dict
         The problem definition
-    X : numpy.ndarray
-        A NumPy array containing the model inputs
-    Y : numpy.ndarray
-        A NumPy array containing the model outputs
-    method : str
-        Type of discrepancy, can be 'CD', 'WD', 'MD' or 'L2-star'.
-        Refer to `scipy.stats.qmc.discrepancy` for more details. Default is WD.
-    print_to_console : bool
+    X, Y : numpy.ndarray
+        An array of model inputs and outputs.
+    method : {"WD", "CD", "MD", "L2-star"}
+        Type of discrepancy. Refer to `scipy.stats.qmc.discrepancy` for more
+        details. Default is "WD".
+    print_to_console : bool, optional
         Print results directly to console (default False)
-    seed : int
+    seed : int, optional
         Seed value to ensure deterministic results
-        Unused, but defined to maintain compatibility.
+        Unused, but defined to maintain compatibility with other functions.
 
     Notes
     -----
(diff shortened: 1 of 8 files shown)


ConnectedSystems:
Thanks @tupui !

cc @samanrazavi

I will take a look when I can over the weekend. Enjoy your Easter break!

tupui:
Thanks you as well 😃

ConnectedSystems:
Very clean implementation as always @tupui .

What would the preferred visualization be for this? Scatter plot with running mean/median?

tupui:
Thanks!

> What would the preferred visualization be for this? Scatter plot with running mean/median?

I think there are 2 options. First is to use the default bar plot as Sobol' and second is to show a grid of canonical plots since this is how we get the indices in the background. Then we could show on each subplot the discrepancy value and do some highlighting of the highest/lowest values. And if we were to add the method we used in the article, then we could show a grid on the subplots and colour/mark the empty cells.

> Actually one other question: how would grouped factors be dealt with?

I am not sure you can do much since you are looking at the input vs output as opposed to just the output for Sobol'. Or maybe merge the values for a given group?? @arnaldpuy?

tupui:
Hey @ConnectedSystems how do you want me to proceed here?

ConnectedSystems:
To support analysis of grouped factors, I suggest simply taking the mean of $S$ for the group. This is what I decided to do for PAWN.

The docstring also needs to be expanded to give an overview of the method. Some hints on interpreting results or pitfalls in analysis would also be appreciated.

tupui:
Sorry for the delay here. I think I addressed your comments.

src/SALib/analyze/discrepancy.py
```diff
@@ -0,0 +1,152 @@
...
+    """Discrepancy indices.
+
+    Parameters
+    ----------
+    problem : dict
+        The problem definition
+    X : numpy.ndarray
+        A NumPy array containing the model inputs
+    Y : numpy.ndarray
+        A NumPy array containing the model outputs
```
ConnectedSystems, line 27: I prefer `np.ndarray` for detailing the argument type, but is specifying the full name the accepted norm? I'll defer to what you say here.

The descriptor holds redundant info as well. I would state what the array holds directly instead, e.g., "An array of model inputs" (the type is already detailed)

src/SALib/analyze/discrepancy.py
```diff
@@ -0,0 +1,152 @@
...
+    is calculated which gives a value for all ``Xi``. This information is used
+    as a sensitivity indice.
+
+    This measure is very fast and is visually explainable. Considering two
+    variables ``X1`` and ``X2``, ``X1`` is more influential than ``X2`` when
+    the scatterplot of ``X1`` against ``Y`` displays a more discernible shape
+    than the scatterplot of ``X2`` against ``Y``.
+
+    For the method to work properly, the input parameter space need to be
+    uniformely covered as the quality of the measure depends on the value of
```
ConnectedSystems, line 52:
```suggestion
    uniformly covered as the quality of the measure depends on the value of
```

src/SALib/analyze/discrepancy.py
```diff
@@ -0,0 +1,152 @@
...
+
+        >>> problem = {
+        ...   'num_vars': 3,
+        ...   'names': ['x1', 'x2', 'x3'],
+        ...   'bounds': [[-np.pi, np.pi]]*3
+        ... }
+        >>> X = latin.sample(problem, 1000)
+        >>> Y = Ishigami.evaluate(X)
+        >>> Si = discrepancy.analyze(problem, X, Y, print_to_console=True)
+
```
ConnectedSystems, line 87: Was the extra empty line necessary for deployed docs?

src/SALib/analyze/discrepancy.py
```diff
@@ -0,0 +1,152 @@
...
+        Unused, but defined to maintain compatibility.
+
+    Notes
+    -----
+    Compatible with:
+        all samplers
+
+    Based on 2D sub projections of ``[Xi,Y]``, the discrepancy of each sample
+    is calculated which gives a value for all ``Xi``. This information is used
+    as a sensitivity indice.
```
ConnectedSystems, line 44:
```suggestion
    as a sensitivity index.
```

Or a slightly more verbose option:


```
    as a measure of sensitivity.
```

ConnectedSystems (changes requested):
Thanks @tupui 

Apologies, I cannot remember what we discussed earlier regarding plotting.
Was the default bar chart sufficient? I think I was intending to add a scatter plot with moving mean like in the pre-print you shared, but not sure if that is necessary or not.

If the bar chart is okay, I think we can merge this after my comments here are considered.

src/SALib/analyze/discrepancy.py
tupui: So we need the full name so that the cross links are working. This is because in `intersphinx_mapping` we use the key `numpy`. We could change this to `np` if you want (and update everywhere, I've seen that we use `numpy.array` sometimes and this should be `numpy.ndarray` at least.)

https://numpydoc.readthedocs.io/en/latest/format.html#parameters

tupui (commented):
Thanks Takuya, I think I addressed your comments :)

tupui:
Hope you don't mind, I just fixed here a few doc warning/issues

src/SALib/analyze/discrepancy.py
ConnectedSystems: Okay, I prefer `np` (personal preference) but I will leave that as a separate task for the future.

ConnectedSystems (approved):

ConnectedSystems:
Thanks @tupui, happy for this to be merged now. I'll leave expanded support for visualization for later.

PR #564, opened 2023-04-06
legend subscript fix for num_vars>10

eyyarbasi:
Legend subscripts were not correctly displayed if `num_vars>10`. This should fix that.

For testing, modify line 14 in _**examples\hdmr\hdmr.py**_ as:
`problem = {"num_vars": 11, "names": ["x1", "x2", "x3","x4","x5","x6","x7","x8","x9","x10","x11"], "bounds": [[-np.pi, np.pi]] * 11}`


ConnectedSystems:
Thank you!

A further modification would be to allow the user to specify which variables to plot (see link below) - I'll make a note to mention this as part of a larger discussion.

https://github.com/SALib/SALib/blob/e5cfb66e3a0879b21afbb427af5188fd661a838b/src/SALib/analyze/rsa.py#L206

Issue #237, opened 2019-04-23
Failing to calculate sensitivity indexes with NAN values in the model results


j-chacon:
I have tried to calculate the Sobol indexes with NAN values as results of the model output, which leads to NAN values of the sensitivity indexes.

A sample code is added

```
import numpy as np
import SALib as sa
from SALib.sample.saltelli import sample
from  SALib.analyze.sobol import analyze

def func(x):
    if x[0]/x[1] > 1:
        out = np.nan
    else:
        out = np.sum(x)
    return out

sa_prob = dict(num_vars = 2,
               names = ['x', 'y'],
               bounds = [[0,1], [0,1]],
               )

X = sample(sa_prob, 1000)
Y = np.array([func(Xi) for Xi in X])
out = analyze(sa_prob, Y)
```
The results of this operation may come from the fact that the functions are not suited to support operations with NAN values (np.sum and np.std among others).

willu47:
Hi @j-chacon. Thanks for adding the issue and an example - very helpful. Have you seen #134 where this is some discussion of missing data?  We may consider SALib failing to produce results when NaN exist to be a feature, rather than a fault...  My worry is that silently producing results when there is missing data may be misleading, whereas the current situation pushes the onus back onto the user to manage their missing data in an appropriate manner.

j-chacon:
Hi @willu47 
Thanks for your answer, indeed a really interesting discussion. 

However, I would like to point out that in some cases, the user may be interested in analysing the sensitivity of a model over a set of feasible combinations of inputs, and not over the whole (fixed) range of the variable. In other words, a non-rectangular domain. In particular, I'm approaching a problem that is quite similar to the one raised in https://github.com/SALib/SALib/issues/134#issuecomment-372028115. In this problem, it is not possible to have a functional relationship among parameters to explore the model.

From the discussion, it is possible to conclude that there may be some approaches such as redo the model runs and interpolate between the results, replacing the NAN value, or completely changing the method (https://github.com/SALib/SALib/issues/134#issuecomment-279782874). 

- Redo the model runs: Sometimes, we are just not interested (or is not physically feasible) that the model can have a certain combination of parameters. This makes the model lie beyond stability, and it would be unrealistic to force it there.
- Interpolate between the results: this approach may be suited in case that the combination of parameters is feasible, but is triggered by instabilities of the model itself. However, does not satisfy every kind of problem.
[...]

Issue #134, opened 2017-02-14
How best to handle model failure with certain parameter sets?


LindsayLBE:
I'm wondering how best to handle cases where some of the saltelli-generated input parameter sets cause model failure, resulting in no output. How will this interfere with the calculating of sensitivity indices using the Sobol method?

willu47:
Hi @LindsayLBE.  This is indeed a problem which I encountered working with the Method of Morris. Check Section 3.4.4.1 (p. 64) of [my thesis](http://discovery.ucl.ac.uk/1504608/) for a brief discussion of this and links to a few sources.

jdherman:
Thanks Will. @LindsayLBE yes it will interfere with the Sobol indices calculations -- it will probably make them return NaN, if I had to guess.

My suggested fixes are just hacks, Will might have better options:
- If you can afford to do the model runs again, try to find the offending values and reduce the parameter ranges so that they don't occur anymore.

- If you can't afford to redo the model runs, and need to just manipulate the data, I'd suggest replacing the missing values with the mean of the other values. This will affect the indices slightly but hopefully not too much.

- The one thing you can't do, at least in the case of Sobol, is remove them completely. The sampling depends on a very specific order of parameter sets that is expected to be preserved. If you were using a different method that works with "given data" (i.e. an arbitrary number and ordering of samples), like the Delta method, then you would be able to just remove these samples and calculate the indices using the remaining ones.

Hope that helps,
Jon

juancastilla:
@jdherman @willu47: I am currently facing a similar problem.

I have a (groundwater) model that may or may not converge depending on parameter combinations that are not entirely trivial i.e., this is not straightforward to fix by simply adjusting the max/min of the prior distributions, there are interaction effects (so hack #1 will not work). I do have control of what the model will output in case of infeasible parameter sets—this can be either NaNs or an unrealistic simulated value that will yield a low likelihood (e.g., sim_value=-9999).

Although the sim_value=-9999/NaN approach can work in the case of MCMC sampling (parameter estimation), I think it can/will corrupt the whole SA. 

[...]

PR #566, opened 2023-04-24
Please revise the code for Sobol and fast

mlmaskey:
These two files contain code that returns NaN values. If I am wrong, please let me know how to solve.
Is there a  way to  upgrade package with updated codes?
Thanks

### DIFF
--- a/src/SALib/analyze/fast.py
+++ b/src/SALib/analyze/fast.py
@@ -104,18 +104,38 @@ def analyze(
     return Si
 
 
-def compute_orders(outputs: np.ndarray, N: int, M: int, omega: int):
-    f = np.fft.fft(outputs)
-    Sp = np.power(np.absolute(f[np.arange(1, math.ceil(N / 2))]) / N, 2)
+# def compute_orders(outputs: np.ndarray, N: int, M: int, omega: int):
+#     f = np.fft.fft(outputs)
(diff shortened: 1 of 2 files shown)


tupui:
Hi @mlmaskey, thank you for reporting.

I am afraid you cannot do that. (Not 100% sure for Fast, as I am not using it.) Sobol' requires a specific set of points to work as intended. If you remove just a single point from the sequence, then it loses its properties. Most importantly in this case, the points are "connected" with each others so if you remove one point, you have to remove the other. Still, this would break the properties of the Sobol' sequence.

What to do here depends on your problem. When you have NaNs, maybe you can replace them with a constant or some other values. Or maybe you have to change the bounds or distributions of your input parameters.

willu47:
There are some useful discussions on missing data in the issue tracker. For example #237 and #134.  I hope those will help you find a solution for your case.

I'm going to close this PR, as SALib should fail to calculate sensitivity indices when there is missing data.  It is up to the user to provide valid data.

tupui:
@willu47 maybe we could add some input validation in general to gracefully error out?

willu47:
Hi @tupui - yes, that's a good idea. Providing some links in the error message to relevant issues where this is discussed could also help. There's probably also a paper to be written on this topic!

tupui:
That's for sure, don't tease me too much 😉

PR #567, opened 2023-05-02
Update JOSS badge to fix Safari issue

tupui:
The current URL is making a lot of re-directions which is preventing the badge from loading correctly on Safari.

Here I am changing the URLs to match latests instructions from JOSS: use the DOI for the target and the JOSS url for the image itself.


ConnectedSystems:
Thanks for the fix @tupui !

Issue #570, opened 2023-06-21
Use reduced-size sample set for bootstrap estimation (delta, sobol)


simonmarwitz:
I use a fairly large pre-generated sample set and use the Delta Moment-Independent Analysis for global sensitivity indices. I want to use only a subset of samples for estimation of delta and S1 indices and the bootstrap estimation of their confidence intervals. Therefore, I have introduced a new Parameter 'Nr' into the routines. Another minor improvement is an switch to deselect additional estimation of first-order sobol indices. As I have trouble creating a fork and a pull request, I attach my code in the hope, that on of you maintainer might merge it into the code base.

```python
from typing import Dict
from scipy.stats import norm, gaussian_kde, rankdata

import numpy as np

from . import common_args
from ..util import read_param_file, ResultDict


def analyze(
    problem: Dict,
    X: np.ndarray,
    Y: np.ndarray,
    num_resamples: int = 100,
    conf_level: float = 0.95,
    print_to_console: bool = False,
    seed: int = None,
    Nr: int = None,
    analyze_sobol: bool = True,
) -> Dict:
    """Perform Delta Moment-Independent Analysis on model outputs.

    Returns a dictionary with keys 'delta', 'delta_conf', 'S1', and 'S1_conf' 
    (first-order sobol indices), where each entry is a list of size D 
    (the number of parameters) containing the indices in the same order as the 
    parameter file. 


    Notes
    -----
    Compatible with:
        all samplers


    Examples
    --------
        >>> X = latin.sample(problem, 1000)
        >>> Y = Ishigami.evaluate(X)
        >>> Si = delta.analyze(problem, X, Y, print_to_console=True)


    Parameters
    ----------
    problem : dict
        The problem definition
    X: numpy.matrix
        A NumPy matrix containing the model inputs
    Y : numpy.array
        A NumPy array containing the model outputs
    num_resamples : int
        The number of resamples when computing confidence intervals (default 100)
    conf_level : float
        The confidence interval level (default 0.95)
    print_to_console : bool
        Print results directly to console (default False)
    Nr: int
        Number of samples to use when resampling (bootstrap) (default None)
    analyze_sobol: int
        Also compute first-order sobol indices and their confidence intervals (default True)


    References
    ----------
    1. Borgonovo, E. (2007). "A new uncertainty importance measure."
           Reliability Engineering & System Safety, 92(6):771-784,
           doi:10.1016/j.ress.2006.04.015.

[...]

PR #571, opened 2023-06-22
Use reduced-size sample set for bootstrap estimation (delta, sobol)

simonmarwitz:
See Issue #570

### DIFF
--- a/src/SALib/analyze/delta.py
+++ b/src/SALib/analyze/delta.py
@@ -54,10 +54,8 @@ def analyze(
     print_to_console : bool
         Print results directly to console (default False)
     y_resamples : int, optional
-
         Number of samples to use when resampling (bootstrap) (default None)
     method : {"all", "delta", "sobol"}, optional
-
         Whether to compute "delta", "sobol" or both ("all") indices (default "all")
 
 
@@ -79,7 +77,7 @@ def analyze(
         y_resamples = Y.size
     
     if not y_resamples <= Y.size:
-        raise ValueError("y_resamples must be less-equal than the total  number of samples")
+        raise ValueError("y_resamples must be less than or equal to the total number of samples")
 
     if not 0 < conf_level < 1:
         raise RuntimeError("Confidence level must be between 0-1.")
@@ -102,7 +100,6 @@ def analyze(
                     Y, Ygrid, X_i, m, num_resamples, conf_level, y_resamples
                 )
             if method in ["all", "sobol"]:
-
                 ind = np.random.randint(Y.size, size=y_resamples)
                 S["S1"][i] = sobol_first(Y[ind], X_i[ind], m)
                 S["S1_conf"][i] = sobol_first_conf(Y, X_i, m, num_resamples, conf_level, y_resamples)
@@ -148,7 +145,6 @@ def calc_delta(Y, Ygrid, X, m):
 
 
 def bias_reduced_delta(Y, Ygrid, X, m, num_resamples, conf_level, y_resamples):
-
     """Plischke et al. 2013 bias reduction technique (eqn 30)"""
     d = np.empty(num_resamples)
 
@@ -186,7 +182,6 @@ def sobol_first(Y, X, m):
 
 
 def sobol_first_conf(Y, X, m, num_resamples, conf_level, y_resamples):
-
     s = np.zeros(num_resamples)
 
     N = len(Y)
--- a/src/SALib/analyze/delta.py
+++ b/src/SALib/analyze/delta.py
@@ -15,12 +15,14 @@ def analyze(
     conf_level: float = 0.95,
     print_to_console: bool = False,
     seed: int = None,
+    Nr: int = None,
 ) -> Dict:
     """Perform Delta Moment-Independent Analysis on model outputs.
 
-    Returns a dictionary with keys 'delta', 'delta_conf', 'S1', and 'S1_conf',
-    where each entry is a list of size D (the number of parameters) containing
-    the indices in the same order as the parameter file.
+    Returns a dictionary with keys 'delta', 'delta_conf', 'S1', and 'S1_conf' 
+    (first-order sobol indices), where each entry is a list of size D 
+    (the number of parameters) containing the indices in the same order as the 
+    parameter file. 
 
 
     Notes
@@ -50,6 +52,8 @@ def analyze(
         The confidence interval level (default 0.95)
     print_to_console : bool
         Print results directly to console (default False)
+    Nr: int
+        Number of samples to use when resampling (bootstrap) (default None)
 
 
     References
@@ -66,15 +70,16 @@ def analyze(
         np.random.seed(seed)
 
     D = problem["num_vars"]
-    N = Y.size
+    if Nr is None:
+        Nr = Y.size
 
     if not 0 < conf_level < 1:
         raise RuntimeError("Confidence level must be between 0-1.")
 
     # equal frequency partition
-    exp = 2.0 / (7.0 + np.tanh((1500.0 - N) / 500.0))
-    M = int(np.round(min(int(np.ceil(N**exp)), 48)))
-    m = np.linspace(0, N, M + 1)
+    exp = 2.0 / (7.0 + np.tanh((1500.0 - Nr) / 500.0))
+    M = int(np.round(min(int(np.ceil(Nr**exp)), 48)))
+    m = np.linspace(0, Nr, M + 1)
     Ygrid = np.linspace(np.min(Y), np.max(Y), 100)
 
     keys = ("delta", "delta_conf", "S1", "S1_conf")
@@ -85,10 +90,11 @@ def analyze(
         for i in range(D):
             X_i = X[:, i]
             S["delta"][i], S["delta_conf"][i] = bias_reduced_delta(
-                Y, Ygrid, X_i, m, num_resamples, conf_level
+                Y, Ygrid, X_i, m, num_resamples, conf_level, Nr
             )
-            S["S1"][i] = sobol_first(Y, X_i, m)
-            S["S1_conf"][i] = sobol_first_conf(Y, X_i, m, num_resamples, conf_level)
+            ind = np.random.randint(Y.size, size=Nr)
+            S["S1"][i] = sobol_first(Y[ind], X_i[ind], m)
+            S["S1_conf"][i] = sobol_first_conf(Y, X_i, m, num_resamples, conf_level, Nr)
     except np.linalg.LinAlgError as e:
         msg = "Singular matrix detected\n"
         msg += "This may be due to the sample size ({}) being too small\n".format(
@@ -130,13 +136,16 @@ def calc_delta(Y, Ygrid, X, m):
     return d_hat
 
 
-def bias_reduced_delta(Y, Ygrid, X, m, num_resamples, conf_level):
+def bias_reduced_delta(Y, Ygrid, X, m, num_resamples, conf_level, Nr):
     """Plischke et al. 2013 bias reduction technique (eqn 30)"""
     d = np.empty(num_resamples)
-    d_hat = calc_delta(Y, Ygrid, X, m)
 
     N = len(Y)
-    r = np.random.randint(N, size=(num_resamples, N))
+    if Nr is None: 
+        Nr = N
+    ind = np.random.randint(N, size=Nr)
+    d_hat = calc_delta(Y[ind], Ygrid, X[ind], m)
+    r = np.random.randint(N, size=(num_resamples, Nr))
     for i in range(num_resamples):
         r_i = r[i, :]
         d[i] = calc_delta(Y[r_i], Ygrid, X[r_i], m)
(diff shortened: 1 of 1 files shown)


src/SALib/analyze/delta.py
```diff
@@ -15,12 +15,15 @@ def analyze(
     conf_level: float = 0.95,
     print_to_console: bool = False,
     seed: int = None,
+    Nr: int = None,
```
tupui, line 18: I propose to follow SciPy's naming here (that would be handy also if we do move towards using `scipy.stats.bootstrap`)

```suggestion
    n_resamples: int = None,

```

src/SALib/analyze/delta.py
```diff
@@ -15,12 +15,15 @@ def analyze(
     conf_level: float = 0.95,
     print_to_console: bool = False,
     seed: int = None,
+    Nr: int = None,
+    analyze_sobol: bool = True,
```
tupui, line 19: I would call it `method` which could get `"all", "sobol", "delta"`. That would be more flexible for future improvements I think.

src/SALib/analyze/delta.py
```diff
@@ -50,6 +53,10 @@ def analyze(
         The confidence interval level (default 0.95)
     print_to_console : bool
         Print results directly to console (default False)
+    Nr: int
```
tupui, line 56: Once we settle on the API, this needs a space in between.
```suggestion
    Nr : int

```

tupui (commented):
Thank you for making a PR @simonmarwitz 🚀 

I left a few suggestions to get this started. We should also probably add some rudimentary tests. Since the changes are fairly "small", I am personally open to leave this for a follow up.

simonmarwitz:
What do I have to do now? I suppose, it is up to me to implement your suggestions and somehow commit them here?!

tupui:
So a typical workflow is that you would then add more commit to the same branch and push, which would update this PR.

This of course provided that @ConnectedSystems agrees with what I am proposing of course 😃 

For collaborative work, it's a bit difficult because I don't have write access to this repo  (yet? 😉). You could give explicit access to people to your fork otherwise. I am happy to guide/help if you need anything.

src/SALib/analyze/delta.py
ConnectedSystems: I think I need someone to explain to me conceptually why we need two separate arguments to do with resampling.

If it is indeed necessary to have both, `n_resamples` is too close to the current `num_resamples`, but I'd like to avoid capitalization in arguments where possible, so agree `Nr` has to be changed...

ConnectedSystems:
Thanks @simonmarwitz , appreciate you contributing even while busy.

Please note my comments on your PR.

@tupui

> I don't have write access to this repo

Just sent you an invite!

src/SALib/analyze/delta.py
tupui: Oh yes I got confused myself as well. `n_resamples` is not a good name here. Sorry for the confusion. I am not sure about naming here, maybe `n_subsamples`?

tupui:
> @tupui
> 
> > I don't have write access to this repo
> 
> Just sent you an invite!

yay thanks Takuya! That's very exciting.

src/SALib/analyze/delta.py
simonmarwitz: If you plan to use scipy.stats.bootstrap in the future, imho it would make sense to adopt the argument names `num_resamples` -> `n_resamples`

I would propose to name the size of the resample `Nr` -> `s_resamples`

src/SALib/analyze/delta.py
simonmarwitz: I propose to completely remove sobol indices from this method and move it to a different file. It is simply not related to DMIM. Although a different implementation of sobol indices already exists in SALib.

src/SALib/analyze/delta.py
tupui: We have a backward compatibility concern here, we sadly cannot rename `num_resamples -> n_resamples` without a deprecation cycle.

`s_resamples`, I am concerned this would be understood as `sobol_resamples`. Since we would reduce Y, why not just `y_resamples` (or `Y_resamples` but I am not a fan of the casing)?

simonmarwitz:
I did some further research on the subject. Apparently here is a paper about it [https://arxiv.org/abs/1112.5016](url). They use resampling without replacement, though. I am not sure if that is easily to be implemented with randint indices. Also `len(Y)` would have to be sufficiently larger than `Nr` / `s_resamples` (what is "sufficiently larger" then?).

src/SALib/analyze/delta.py
simonmarwitz: If you go for my proposal to remove sobol indices from this method, that would no longer be a problem.

tupui: Please don't remove Sobol' indices from this method. We cannot do this like that for backward compatibility reasons. This is going to break people's code.

src/SALib/analyze/delta.py
tupui: I replied above. But in the literature you do see expressions of Sobol' indices with moment independent. So this is just a different metric. I do agree that the name of the function is not general enough then to convey that. But we cannot change like this the parameters and features.

What we can do though is (in some follow up) to think about whether we would want or not to have a new API that would be more general. Doing so, we would be able to rework the API as we want and then ultimately deprecate this function.

src/SALib/analyze/delta.py
ConnectedSystems: I'm okay with `y_resamples`, seems like a good interim compromise. 

I'm also open to removing Sobol' (not in this PR, but in the future).

src/SALib/analyze/delta.py
```diff
@@ -130,13 +144,16 @@ def calc_delta(Y, Ygrid, X, m):
     return d_hat
 
 
-def bias_reduced_delta(Y, Ygrid, X, m, num_resamples, conf_level):
+def bias_reduced_delta(Y, Ygrid, X, m, num_resamples, conf_level, Nr):
```
tupui, line 147:
```suggestion
def bias_reduced_delta(Y, Ygrid, X, m, num_resamples, conf_level, y_resamples):

```

src/SALib/analyze/delta.py
```diff
@@ -130,13 +144,16 @@ def calc_delta(Y, Ygrid, X, m):
...
     d = np.empty(num_resamples)
-    d_hat = calc_delta(Y, Ygrid, X, m)
 
     N = len(Y)
-    r = np.random.randint(N, size=(num_resamples, N))
+    if Nr is None: 
+        Nr = N
+    ind = np.random.randint(N, size=Nr)
+    d_hat = calc_delta(Y[ind], Ygrid, X[ind], m)
+    r = np.random.randint(N, size=(num_resamples, Nr))
```
tupui, line 156: I suppose we don't need to handle the None case as this function is more or less private.
```suggestion
    ind = np.random.randint(N, size= y_resamples)
    d_hat = calc_delta(Y[ind], Ygrid, X[ind], m)
    r = np.random.randint(N, size=(num_resamples, y_resamples))

```

src/SALib/analyze/delta.py
```diff
@@ -165,11 +182,13 @@ def sobol_first(Y, X, m):
     return Vi / np.var(Y)
 
 
-def sobol_first_conf(Y, X, m, num_resamples, conf_level):
+def sobol_first_conf(Y, X, m, num_resamples, conf_level, Nr):
```
tupui, line 185:
```suggestion
def sobol_first_conf(Y, X, m, num_resamples, conf_level, y_resamples):

```

src/SALib/analyze/delta.py
```diff
@@ -165,11 +182,13 @@ def sobol_first(Y, X, m):
...
 
-def sobol_first_conf(Y, X, m, num_resamples, conf_level):
+def sobol_first_conf(Y, X, m, num_resamples, conf_level, Nr):
     s = np.zeros(num_resamples)
 
     N = len(Y)
-    r = np.random.randint(N, size=(num_resamples, N))
+    if Nr is None: 
+        Nr = N
+    r = np.random.randint(N, size=(num_resamples, Nr))
```
tupui, line 191: Same here
```suggestion
    r = np.random.randint(N, size=(num_resamples, y_resamples))

```

src/SALib/analyze/delta.py
```diff
@@ -50,6 +53,10 @@ def analyze(
         The confidence interval level (default 0.95)
     print_to_console : bool
         Print results directly to console (default False)
+    y_resamples : int
+        Number of samples to use when resampling (bootstrap) (default None)
+    method : str
```
tupui, line 58:
```suggestion
    method : {"all", "delta", "sobol"}, optional

```

src/SALib/analyze/delta.py
```diff
@@ -50,6 +53,10 @@ def analyze(
         The confidence interval level (default 0.95)
     print_to_console : bool
         Print results directly to console (default False)
+    y_resamples : int
```
tupui, line 56:
```suggestion
    y_resamples : int, optional

```

src/SALib/analyze/delta.py
```diff
@@ -84,11 +95,14 @@ def analyze(
...
-            S["delta"][i], S["delta_conf"][i] = bias_reduced_delta(
-                Y, Ygrid, X_i, m, num_resamples, conf_level
-            )
-            S["S1"][i] = sobol_first(Y, X_i, m)
-            S["S1_conf"][i] = sobol_first_conf(Y, X_i, m, num_resamples, conf_level)
+            if method in ["all", "delta"]:
+                S["delta"][i], S["delta_conf"][i] = bias_reduced_delta(
+                    Y, Ygrid, X_i, m, num_resamples, conf_level, y_resamples
+                )
+            if method in ["all","sobol"]:
```
tupui, line 102:
```suggestion
            if method in ["all", "sobol"]:

```

tupui (changes requested):
Thanks for the update! This looks in good shape besides a few tests that need to be updated. (I made suggestions to help 😃, you can go to the file tabs and group suggestions together.)

ConnectedSystems:
I'll try to finalize comments/suggestions this week. Sorry for the delay in getting to this!

src/SALib/analyze/delta.py
ConnectedSystems: On second thought, I think best to leave the Sobol' estimation in (but optional).

I prefer `y_resamples`.

src/SALib/analyze/delta.py
```diff
@@ -50,6 +53,12 @@ def analyze(
         The confidence interval level (default 0.95)
     print_to_console : bool
         Print results directly to console (default False)
+    y_resamples : int, optional
+
+        Number of samples to use when resampling (bootstrap) (default None)
```
ConnectedSystems, line 58: Could you make the spacing here consistent?

src/SALib/analyze/delta.py
```diff
@@ -66,15 +75,19 @@ def analyze(
         np.random.seed(seed)
 
     D = problem["num_vars"]
-    N = Y.size
+    if y_resamples is None:
+        y_resamples = Y.size
+    
+    if not y_resamples <= Y.size:
+        raise ValueError("y_resamples must be less-equal than the total  number of samples")
```
ConnectedSystems, line 82: Remove extra space in error message ("total  number")

"less-equal" -> "less than or equal to"

src/SALib/analyze/delta.py
```diff
@@ -84,11 +97,15 @@ def analyze(
...
-                Y, Ygrid, X_i, m, num_resamples, conf_level
-            )
-            S["S1"][i] = sobol_first(Y, X_i, m)
-            S["S1_conf"][i] = sobol_first_conf(Y, X_i, m, num_resamples, conf_level)
+            if method in ["all", "delta"]:
+                S["delta"][i], S["delta_conf"][i] = bias_reduced_delta(
+                    Y, Ygrid, X_i, m, num_resamples, conf_level, y_resamples
+                )
+            if method in ["all", "sobol"]:
+
```
ConnectedSystems, line 105: Remove blank line here please

src/SALib/analyze/delta.py
```diff
@@ -130,13 +147,16 @@ def calc_delta(Y, Ygrid, X, m):
     return d_hat
 
 
-def bias_reduced_delta(Y, Ygrid, X, m, num_resamples, conf_level):
+def bias_reduced_delta(Y, Ygrid, X, m, num_resamples, conf_level, y_resamples):
+
```
ConnectedSystems, line 151: Please remove extra line here

src/SALib/analyze/delta.py
```diff
@@ -165,11 +185,13 @@ def sobol_first(Y, X, m):
     return Vi / np.var(Y)
 
 
-def sobol_first_conf(Y, X, m, num_resamples, conf_level):
+def sobol_first_conf(Y, X, m, num_resamples, conf_level, y_resamples):
+
```
ConnectedSystems, line 189: Same here, remove extra line

ConnectedSystems (changes requested):
Hi @simonmarwitz 

Thanks for these changes, I have a few more minor requests before I approve and merge.

Thanks again!

ConnectedSystems (approved):
Thanks @simonmarwitz !

ConnectedSystems:
Hmm, some tests failing - I'll look into it soon.

ConnectedSystems:
Apologies all, not sure what happened there as I definitely didn't mean to merge but the test fix is small.

PR #572, opened 2023-06-23
Update landing page

tupui:
Fix links and update the list of methods.

I am also hopping that adding some spaces around the code blocks would allow them to be rendered properly. The preview of the markdown looks ok.

### DIFF
--- a/index.md
+++ b/index.md
@@ -6,15 +6,12 @@ Python implementations of commonly used sensitivity analysis methods. Useful in 
 
 **Requirements:** [NumPy](http://www.numpy.org/), [SciPy](http://www.scipy.org/), [matplotlib](http://matplotlib.org/)
 
-**Installation:** `pip install SALib` or `python setup.py install`
+**Installation:** ``pip install SALib`` or ``pip install .`` or ``conda install SALib``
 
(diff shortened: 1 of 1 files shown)


ConnectedSystems:
Hi @tupui, this looks like its for the github pages (http://salib.github.io/SALib/) - is that correct?

I have half a mind to remove it given we link to ReadTheDocs in the readme, and it constitutes a (small) duplication and maintenance burden.

@willu47 any thoughts here?

I'll accept the changes for now, however.

tupui:
I agree that as it is there is not much value in having this page. I would also remove it. Only thing to consider is that it might break some links if people are using this. I don't know here if that's relevant.

For the CI. I think we can ignore since this does not need to be built on RTD. It would just be GH rendering the readme.

willu47:
Hi. I would be happy removing this landing page, but I'd suggest setting up forwarding to the read-the-docs so as not to break to the URL.  I know some citations that reference this URL, and not the GitHub repo, for example.

tupui:
Ok I will make another PR with the redirect 👍 (I got to the same link before you edited your comment 😅)

ConnectedSystems:
Closing as this is addressed by #574

PR #574, opened 2023-06-26
Redirection to https://salib.readthedocs.io

tupui:
Superseed #572

Locally does the proper redirection in both Chrome and Safari.

I left the config file because already there. Although apparently it's not needed anymore.


ConnectedSystems (approved):
Thanks! All LGTM

PR #575, opened 2023-06-28
Enhanced HDMR Method

sahin-abdullah:
The first implementation of HDMR was an approximation method which means that decomposing the model output to its inputs may not be accurate for all models. It is pretty accurate though when inputs are uncorrelated. However, most of the real-world models have correlation among input parameters. Model output's decomposition to its inputs is outlined in Sobol's paper as below.

```math
\tilde{y} \approx \widehat{y} = f_0 + \sum_{i=1}^{d} f_i(x_i) + 
              \sum_{i=1}^{d-1} \sum_{j=i+1}^{d} f_{ij} (x_{ij}) + 
              \sum_{i=1}^{d-2} \sum_{j=i+1}^{d-1} 
              \sum_{j+1}^{d} f_{ijk} (x_{ijk}) + \epsilon 
```

It was pointed out that components (a.k.a. summands used by Sobol) has to be orthogonal to each other and it is the condition needs to be addressed during the decomposition. However, when input parameters are correlated, decomposition is not with orthogonality condition. Later on, Hooker(2007) came with hierarchical orthogonal condition that allows unique decomposition. It basically says that components do not need to be orthogonal to each other, they should be orthogonal hierarchically. For example, the component $f_{123}$ should be orthogonal to its lower order components, $f_{0}, f_{1}, f_{2}, f_{3}, f_{12}, f_{13}, f_{23}$. Also, this condition is also valid when there is no correlation between parameters.  

With this unique decomposition, one can account for correlated contribution to the output variance of the model output.

There are several items that enhanced current implementation.

1. We keep original implementation with changing the basis matrix only. Earlier implementation was using cubic b-splines, now we are using orthonormal polynomials which construct better basis matrix. `extended_base` key of `options` should be set to `False` to use the original implementation.
2. New extended base approach. This provides unique decomposition and hence better sensitivity indices.
3. Method parameters are more self-explanatory. Instead of using K, R, N etc., we are using bootstrap, subset, l2_penalty etc.
4. Removed redundant variables and computations. `_init()` method is replaced by `_core_params()` method and unnecessary variables are removed
5. Added docstring to all methods
6. Emulator part. Previous implementation was returning extra variables due to emulator and it was plotting emulator performance. In this version, emulator handled in a different way to make it concise and we are not plotting any figure to see its performance. One can calculate the SSR or any other related metric. 

It took a lot of effort to write one single `_basis_matrix()` method to account for both extended base and without extended base approach. I hope it will be helpful to the researchers who take advantage of SALib.

I will add the test cases later on.


ConnectedSystems:
@sahin-abdullah would you mind please creating a branch in your fork for this PR? You've submitted this PR from your `main` and it seems to be confusing my git client.

PR #580, opened 2023-07-06
Scaled elementary effects - method of Morris

NMoksnes:
For the method of Morris calculations, @willu47 and I have added an additional calculation step based on [Sin and Gernaey (2009)](https://doi.org/10.1016/S1570-7946(09)70154-3) for scaling the elementary effect, allowing for comparison between different model outputs.

The default setting is to calculate the original Morris calculations, with the same input and output as previous versions. 

To retrieve the scaled elementary effects the matrix **X** needs to be the actual steps of the input parameters in a numpy array.
Furthermore, the `scaled ` parameter needs to be set to `True`.

`analyze_morris.analyze(problem, X, Y, scaled=True)`

The scaled elementary effects are calculated per trajectory as proposed in [Moret et al. (2017)](https://doi.org/10.1016/j.apenergy.2017.05.106) and return the same ranking of the input parameters as the elementary effects.

Two tests have also been provided with this pull request.

This also corrects a mistake in the code, where the `mu` metric is computed for groups using the incorrect function. We replace `Si["mu"] = _compute_grouped_sigma(mu, groups)` with `Si["mu"] = _compute_grouped_metric(mu, groups)`


ConnectedSystems:
Thanks @NMoksnes @willu47 

I will aim to do a review sometime this week. I've been preparing for/attending a conference, hence the delay in getting to this.

src/SALib/analyze/morris.py
```diff
@@ -31,9 +32,17 @@ def analyze(
...
     - ``sigma`` is the standard deviation of the distribution
 
+    When ``scaled`` is True, the elementary effects are scaled by the ratio of
+    standard deviation of ``X`` and ``Y`` according to [3]. When using this
+    option it is important to ensure that ``X`` contains the actual values
+    passed into the model, as the elementary effects are divided by the
+    step calculated from ``X`` rather than using `delta` which is calculated from
+    the number of levels used in the sample. This could be the case if you
+    perform post-processing on the values before passing them to the model.
+
```
ConnectedSystems, line 45: It would also be good to add some information on when it is desirable to use the scaled, non-dimensional, sensitivity measure (e.g., avoids the dependency between factor rankings and the units of the factors)

src/SALib/analyze/morris.py
```diff
@@ -280,9 +311,10 @@ def _reorganize_output_matrix(
     increase: bool
         Direction to consider (values that increased or decreased). "Increase"
         is the default value.
+
     Returns
     -------
-
+    numpy.ndarray
```
ConnectedSystems, line 317:
```suggestion
    np.ndarray
```

src/SALib/analyze/morris.py
```diff
@@ -304,6 +336,7 @@ def _compute_elementary_effects(
     model_outputs: np.ndarray,
     trajectory_size: int,
     delta: float,
+    scaling=False,
```
ConnectedSystems, line 339: Add type hint

src/SALib/analyze/morris.py
```diff
@@ -319,11 +352,25 @@ def _compute_elementary_effects(
...
     elementary_effects : np.array
         Elementary effects for each parameter
+
+    References
+    ----------
+    1. Sin and Gearney (2009)
+       Improving the Morris Method for Sensitivity Analysis
+       by Scaling the Elementary Effects.
+       19th European Symposium on Computer Aided Process Engineering ESCAPE19:925-930
+    2. Moret et al. (2017)
```
ConnectedSystems, line 369: From memory, a space between the two list items are needed for the list to render correctly in the deployed documentation.

src/SALib/analyze/morris.py
```diff
@@ -345,8 +392,32 @@ def _compute_elementary_effects(
...
+            sigma_x_per_trajectory,
+            sigma_y_per_trajectory[:, np.newaxis],
+            out=np.zeros_like(sigma_x_per_trajectory),
+            where=mask,
+        )
+        elementary_effects = see_dy_dx * adjustment_trajectory.T
+
+    else:
+
+        elementary_effects = np.divide(difference, delta)
```
ConnectedSystems, line 420: Is use of `np.divide` required? (as opposed to `difference / delta`)

ConnectedSystems (changes requested):
Thank you @NMoksnes @willu47 , looks like a very clean update.

Some minor comments for your consideration.

src/SALib/analyze/morris.py
ConnectedSystems: Actually, @tupui what's the norm for these? Should we be specifying the package name fully?

ConnectedSystems:
Hi @willu47 @NMoksnes I realised this morning that adjustments will be made to support use of `scaled` through the command-line interface. Let me know if you need help with this.

src/SALib/analyze/morris.py
NMoksnes: No, I think you are correct as delta is a float in this case. This is the original formulation,

NMoksnes:
Thank you for your review @ConnectedSystems and for your valuable suggestions. I have accepted all of them and also added the additional information requested. As for the division of delta for the EE, I leave that decision to you. 

For the command line interface if you or @willu47 have the time to propose the changes needed then I would be very grateful.

Nandi

NMoksnes:
Thank you, @willu47 for modifying the command line interface.

src/SALib/analyze/morris.py
ConnectedSystems: In this case, please remove use of `np.divide`.

The `/` is shorthand for `np.divide` anyway. 
https://numpy.org/doc/stable/reference/generated/numpy.divide.html

If you're interested in the technical details, on top of the above Python has a constant function call overhead - meaning just calling a function (any function) adds $x$ amount of time. Admittedly it is a small amount of time, so it's a micro-optimisation but I usually try to err on the side of simplicity :)

image

NMoksnes:
I think all comments have now been covered from your review @ConnectedSystems. Let me know if there is anything else needed.

ConnectedSystems (approved):

PR #591, opened 2023-10-13
Standardize line terminations

musicinmybrain:
Convert the minority of files that are CRLF-terminated (Windows/DOS style) to LF-terminated (POSIX-style).

This can be replicated with:

```sh
find * -type f -exec file '{}' + |
  awk -F : '/CRLF/ { print $1 }' |
  xargs dos2unix
```


ConnectedSystems:
Thanks for these PRs and helping standardise everything!

PR #597, opened 2023-11-08
loguniform scaling

joergbuchwald:
add a loguniform scaling of samples

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -220,7 +220,6 @@ def test_nonuniform_scale_samples_logunif():
     rng = np.random.default_rng()
     x = stats.loguniform.rvs(size=100000, a=1e-22, b=1e-18, random_state=rng)
     res = stats.cramervonmises_2samp(actual[:, 0], x)
-    alpha = 0.05
     zero_hypothesis_rejected = True
     if res.pvalue > 0.05:
         zero_hypothesis_rejected = False


tupui (commented):
Thank you @joergbuchwald.

I am really not sure about growing this list as there are more than 100 distributions in SciPy. It will be more and more difficult to justify the inclusion or not of specific distributions.

Instead I think we should make the infra more flexible as to accept a distribution object or use the latest sampling features we added in SciPy (the work on UNURAN).

We are actually thinking about making an overall of the API.

Still, in the meantime, this has some value and I would be ok integrating the changes (modulo a few things we can discuss if we do continue here).

@ConnectedSystems what do you think?

ConnectedSystems:
I think in the short term we can accept this.

1. The work has already been done
2. Others will benefit from it
3. Any changes to move to a more flexible infrastructure will take some time

For the above reasons, I think accepting this PR is the beneficial path forward 😄 

My only comment is that some docs should be added, which I can do if needed.

src/SALib/util/__init__.py
```diff
@@ -205,6 +206,15 @@ def _nonuniform_scale_samples(params, bounds, dists):
             else:
                 conv_params[:, i] = params[:, i] * (b2 - b1) + b1
 
+        elif dists[i] == "logunif":
+            if b1 >= b2:
```
tupui, line 210: Not sure this is needed. The SciPy underlying function function would raise. But ok if we do want a specific messaging.

Thought for later: we should let errors bubble up and catch all if we want to have a personalized message instead.

tests/test_util.py
```diff
@@ -203,3 +203,31 @@ def test_nonuniform_scale_samples_truncnorm():
...
+            [3.03290952e-20],
+            [1.04680308e-21],
+            [5.09015386e-19],
+            [2.36662233e-21],
+            [1.11611357e-19],
+            [1.35560739e-20],
+            [5.86336757e-21],
+        ]
+    )
+    np.testing.assert_allclose(actual, expected)
```
tupui, line 233: A better test would be to check that the sample does follow the prescribed distribution. This can be done for instance with a fit (and compare the fitted params) or even a classical 2 sample test like Cramer or KS. This would be a more robust test and would not need seeding even.

tupui (commented):
Ok then 👍 I have some minor review 😃

joergbuchwald:
> I think in the short term we can accept this.
> 
>     1. The work has already been done
> 
>     2. Others will benefit from it
> 
>     3. Any changes to move to a more flexible infrastructure will take some time
> 
> 
> For the above reasons, I think accepting this PR is the beneficial path forward 😄
> 
> My only comment is that some docs should be added, which I can do if needed.

Is there a specific place for the docs? I think I found the parametrization for the other distributions in the source code only :laughing:

joergbuchwald:
> Thank you @joergbuchwald.
> 
> I am really not sure about growing this list as there are more than 100 distributions in SciPy. It will be more and more difficult to justify the inclusion or not of specific distributions.
> 
> Instead I think we should make the infra more flexible as to accept a distribution object or use the latest sampling features we added in SciPy (the work on UNURAN).
> 
> We are actually thinking about making an overall of the API.
> 
> Still, in the meantime, this has some value and I would be ok integrating the changes (modulo a few things we can discuss if we do continue here).
> 
> @ConnectedSystems what do you think?

I agree, a more flexible general way to access scipy would be nice. However, it would break the current parametrization/API. For the time being, I just added what I needed for the moment.

ConnectedSystems:
> Is there a specific place for the docs? I think I found the parametrization for the other distributions in the source code only 😆

There's a bit of duplication happening but if you add something here we can consolidate and reorganize later :) thank you!

https://github.com/SALib/SALib/blob/fef0cd6ce851d271acf01259d225489f5b3847fc/docs/user_guide/advanced.rst?plain=1#L113

joergbuchwald:
There is now a non-zero chance that the test will fail.

tests/test_util.py
```diff
@@ -203,3 +204,24 @@ def test_nonuniform_scale_samples_truncnorm():
...
+    }
+    actual = latin.sample(problem, 10000)
+    rng = np.random.default_rng()
+    x = stats.loguniform.rvs(size=100000, a=1e-22, b =1e-18, random_state=rng)
+    res = stats.cramervonmises_2samp(actual, x)
+    alpha = 0.05
+    zero_hypothesis_rejected = True
+    if res.pvalue > 0.05:
+        zero_hypothesis_rejected = False
+    np.testing.assert_equal(zero_hypothesis_rejected, False)
```
tupui, line 227:
```suggestion
    assert zero_hypothesis_rejected is False
```

docs/user_guide/advanced.rst
```diff
@@ -110,6 +111,7 @@ parameter bounds but sample-specific metadata.
      A soon-to-be deprecated two-value format assumes the lower bound to be 0
      e.g. :code:`[3, 0.5]` assumes 0 to 3, with a peak at 1.5
 * norm: normal distribution with mean and standard deviation
+* truncnorm: runcated normal distribution with upper and lower bounds, mean and standard deviation
```
tupui, line 114:
```suggestion
* truncnorm: truncated normal distribution with upper and lower bounds, mean and standard deviation
```

tupui (approved):
Great thank you 👍 just a minor suggestion and a typo from me.

tests/test_util.py
```diff
@@ -203,3 +204,24 @@ def test_nonuniform_scale_samples_truncnorm():
...
+    problem = {
+        "num_vars": 1,
+        "dists": ["logunif"],
+        "bounds": [[1e-22, 1e-18]],
+        "names": ["x1"],
+    }
+    actual = latin.sample(problem, 10000)
+    rng = np.random.default_rng()
+    x = stats.loguniform.rvs(size=100000, a=1e-22, b =1e-18, random_state=rng)
+    res = stats.cramervonmises_2samp(actual, x)
```
tupui, line 222: It looks like this is failing in the tests

tests/test_util.py
```diff
@@ -203,3 +204,24 @@ def test_nonuniform_scale_samples_truncnorm():
...
+    problem = {
+        "num_vars": 1,
+        "dists": ["logunif"],
+        "bounds": [[1e-22, 1e-18]],
+        "names": ["x1"],
+    }
+    actual = latin.sample(problem, 10000)
+    rng = np.random.default_rng()
+    x = stats.loguniform.rvs(size=100000, a=1e-22, b=1e-18, random_state=rng)
+    res = stats.cramervonmises_2samp(actual[:,0], x)
```
tupui, line 222:
```suggestion
    res = stats.cramervonmises_2samp(actual[:, 0], x)
```

tupui:
FYI, there is no need to force push. We even recommend not to do that because it breaks GitHub's links.

tests/test_util.py
```diff
@@ -203,3 +204,24 @@ def test_nonuniform_scale_samples_truncnorm():
...
+        "dists": ["logunif"],
+        "bounds": [[1e-22, 1e-18]],
+        "names": ["x1"],
+    }
+    actual = latin.sample(problem, 10000)
+    rng = np.random.default_rng()
+    x = stats.loguniform.rvs(size=100000, a=1e-22, b=1e-18, random_state=rng)
+    res = stats.cramervonmises_2samp(actual[:, 0], x)
+    alpha = 0.05
+    zero_hypothesis_rejected = True
```
tupui, line 224:
```suggestion
    res = stats.cramervonmises_2samp(actual[:, 0], x)
    zero_hypothesis_rejected = True
```

tupui:
The failures in CI are not related. @ConnectedSystems this is ready on my side.

ConnectedSystems:
Awesome! Thank you @joergbuchwald

PR #609, opened 2024-03-03
Update dependencies

tupui:
This PR

- Bump minimal Python version
- Update deps
- Use hatch-vcs to simplify the release process. Instead people can access all metadata with `importlib.metadata.version("SALib")`. See https://docs.python.org/3/library/importlib.metadata.html


tupui:
Ok the CI is also fixed for mamba updating Python. Before we had the same version running on both jobs.

tupui:
Ooops will revert the change for direction...

tupui:
@ConnectedSystems this is in good shape now, madness is fixed 😄

ConnectedSystems:
> Ok the CI is also fixed for mamba updating Python. Before we had the same version running on both jobs.

Wow, thanks for catching that.

Everything looks good

tupui:
Thanks :)

PR #617, opened 2024-04-22
Add `base_sequence` to `sample.sobol`

tupui:
I had to do this for a project to change the sampling method.

I am not sure about this and maybe we should start thinking about using the new SciPy machinery.

What do you think?

### DIFF
--- a/src/SALib/sample/sobol.py
+++ b/src/SALib/sample/sobol.py
@@ -15,6 +15,7 @@ def sample(
     calc_second_order: bool = True,
     scramble: bool = True,
     skip_values: int = 0,
+    base_sequence: Optional[np.ndarray] = None,
     seed: Optional[Union[int, np.random.Generator]] = None,
 ):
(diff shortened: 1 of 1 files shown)


ConnectedSystems:
You mean replace this implementation with what is in SciPy completely?

tupui:
Yeah to see how the new functions can be used on our side. Same goes for Sobol' indices themselves

ConnectedSystems:
Sorry for the long delay in response.

I wonder if we can make the `base_sequence` also accept a function that, if given a target shape, will generate the array accordingly:

```python
target_shape = (2 * D, N)
seq = base_sequence(target_shape)
```

Just a thought...

Otherwise I am fine with the current implementation

tupui:
I see what you mean. I am not sure, do you have a use case yourself? This param is already questionable 😅 I don't mind adding though if you want. At some point we might either way to revisit the whole API (sad CZI 😓)

PR #628, opened 2024-07-14
Updates for numpy 2.0

### LINKED ISSUE #625: Migrate to numpy>=2.0.0
Numpy 2.0.0 was released at June 16. It seems that SALib is incompatible with the new numpy now.

E.g., the quick start example, if run this after `pip install SALib --force-reinstall`
```python
from SALib.test_functions import Ishigami
import numpy as np

problem = {
  'num_vars': 3,
  'names': ['x1', 'x2', 'x3'],
  'bounds': [[-np.pi, np.pi]]*3
}

# Generate samples
param_values = saltelli.sample(problem, 1024)

# Run model (example)
Y = Ishigami.evaluate(param_values)

# Perform analysis
Si = sobol.analyze(problem, Y, print_to_console=True)
# Returns a dictionary with keys 'S1', 'S1_conf', 'ST', and 'ST_conf'
# (first and total-order indices with bootstrap confidence intervals)
```

will yield the following error message,
```
Traceback (most recent call last):
  File "...", line 19, in <module>
    Si = sobol.analyze(problem, Y, print_to_console=True)
  File "python3.9/site-packages/SALib/analyze/sobol.py", line 151, in analyze
    S["S1"][j] = first_order(A, AB[:, j], B)
  File ".../python3.9/site-packages/SALib/analyze/sobol.py", line 215, in first_order
    if y.ptp() == 0:
AttributeError: `ptp` was removed from the ndarray class in NumPy 2.0. Use np.ptp(arr, ...) instead.
```

Any plan on migrating to numpy>=2.0.0?

ConnectedSystems:
Removes use of deprecated/removed numpy methods to be compatible with numpy 2.0

Closes #625

### DIFF
--- a/src/SALib/test_functions/Sobol_G.py
+++ b/src/SALib/test_functions/Sobol_G.py
@@ -81,7 +81,7 @@ def evaluate(values, a=None, delta=None, alpha=None):
         mod_x = shift_of_x - integral
         temp_y = np.abs(2 * mod_x - 1) ** alpha
         y_elements = ((1 + alpha) * temp_y + a) / (1 + a)
-        Y[i] = np.prod(y_elements)
+        Y[i] = y_elements.prod()
 
     return Y
 
@@ -103,7 +103,7 @@ def _total_variance(a=None, alpha=None):
         alpha = np.ones_like(a)
 
     a = np.array(a)
-    return np.add(-1, np.product(1 + _partial_first_order_variance(a, alpha), axis=0))
+    return np.add(-1, np.prod(1 + _partial_first_order_variance(a, alpha), axis=0))
 
 
 def sensitivity_index(a, alpha=None):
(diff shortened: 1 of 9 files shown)


src/SALib/test_functions/Sobol_G.py
```diff
@@ -81,7 +81,7 @@ def evaluate(values, a=None, delta=None, alpha=None):
         mod_x = shift_of_x - integral
         temp_y = np.abs(2 * mod_x - 1) ** alpha
         y_elements = ((1 + alpha) * temp_y + a) / (1 + a)
-        Y[i] = np.prod(y_elements)
+        Y[i] = y_elements.prod()
```
tupui, line 84: This one is a mistake no?

tupui (commented):
Thanks @ConnectedSystems you beat me to it 🚀 

Just not sure about one change.

src/SALib/test_functions/Sobol_G.py
ConnectedSystems: It was intentional as a micro-performance change.

Calling `np.prod` is slightly slower than `array.prod()` due to how Python looks up package functions vs methods.

I realise it's technically out-of-scope for this PR, but I figured it's a one line change... so I might as well do it now while I was at it.

tupui (approved):
Ok thanks 👍

If we were to add array API support we would revisit but good for now.

PR #638, opened 2024-09-09
Add standard deviation to reported sensitivity statistic

ConnectedSystems:
Long overdue update - add the standard deviation as one of the reported summary stats.

The standard deviation (`stdev`) is added at the end of the list to minimize the chance that this will break existing code (e.g., extracting the `CV` value by position).


tupui (approved):
Fine by me 😃

PR #640, opened 2024-09-20
Updated Numpy dependency to `>=2.0`

### LINKED ISSUE #639: Conda dependency issue
Hi SALib devs!

There seems to be a dependency issue in your library when installing through Conda right now. You've pinned Numpy support at `numpy>=1.20.3` ([conda forge feedstock](https://github.com/conda-forge/salib-feedstock/blob/b38d5e435b42661ce027fb3d17a098391306d240/recipe/meta.yaml#L21C7-L21C21)) but with your newest release `v1.5.1` you've dropped support for `numpy<2.0`. For example here:

https://github.com/SALib/SALib/blob/942a2d4c40c73e868abe464d6b21763820c0dc9c/src/SALib/analyze/delta.py#L146

The `np.trapezoid` function doesn't exist pre Numpy 2.0, causing execution to fail.

Could you rebuild for Conda using a new pin?

Thanks a lot!

mrvisscher:
Updated Numpy dependency to `>=2.0` because of the issue described in #639 where SALib execution fails because support for `Numpy<2.0` has been dropped.

* Closes #639 

**Note:** this only fixes the dependency issue for Conda, not for PyPI

### Related PR:
* https://github.com/conda-forge/salib-feedstock/pull/41

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
 dependencies:
   # required
-  - numpy>=1.20.3
+  - numpy>=2.0
   - pandas>=2.0
   - matplotlib>=3.5
   - scipy>=1.9.3


ConnectedSystems:
Thanks for the PR @mrvisscher 

@tupui yes, I am okay with requiring numpy v2.0 now

PR #645, opened 2025-01-03
Update morris.py

AnbazhaganS:
Typo in default value of num_resamples. Documentation says its default value is 1000 when the default is 100 in the code.


tupui (changes requested):
Thanks @AnbazhaganS for the PR.

We should instead change the documentation otherwise that is a breaking change as users could rely on this default value.

tupui (approved):
Thanks 🚀

PR #652, opened 2025-03-16
Fix FAST CI estimation via bootstrapping from contiguous blocks

### LINKED ISSUE #649: Possible wrong method in the Bootstrapping of the eFAST analyzis
Hi, after running some tests on the eFAST model I found odd behaviour of the generated Confidence Intervals.

I tested the model

$$f(x_1,x_2) = x_1 a_1 + x_2 a_2 $$

for varying a_1.

Currently, the $CI(S_i)$ show no correlation to the size of $S_i$

Image

this is due to the random sampling done [here](https://github.com/SALib/SALib/blob/main/src/SALib/analyze/fast.py#133) as:

`
sample_idx = np.random.choice(T_data, replace=True, size=n_size)
`

This destroys the function incorporated in the model ouput.

Unfortunately, I do not know what the correct bootstrapping method would be. Two methods that could work would be:

1. sorting after the random sampling
2. using a smaller sample but retaining order and position (e.g. sample_idx = linspace(random_int,random_int + n_size) ... of course some overflow consideration is needed)

These Methods result in:

ImageImage

For comparison, I did the same analysis with Sobol and a kstest. My kstests implementation is consistent with the results of method 2 and the sobol implementation in this library suggests method 1. What are your Thoughts on this ?

Image

ConnectedSystems:
Correct the bootstrapping method applied for FAST analysis by replacing the sampling with replacement approach with a block boostrapping approach which, in my view, better preserves the local frequency structure.

Closes #649 (thanks to @SchnFlo for suggesting the fix).

In #649, two potential methods are proposed to resolve the bug.

The first (method 1) is to first sort the samples, producing results that were reportedly more in line with the Sobol' confidence bounds.
The second (method 2) was to apply the block boostrapping approach.

Ultimately, I decided to go with the block bootstrap (method 2) as I think it aligns more with the FAST analysis method (which is based on frequency analysis).


ConnectedSystems:
Tests are failing due to an outdated workflow config.

#653 has the necessary changes to resolve the issue so once that's merged in I'll rebase this PR.

ConnectedSystems:
While I updated the approach, ultimately I think we should deprecate support for bootstrapping as it is inappropriate for FAST-based analyses (a multi phase-shift approach should be adopted instead). I've elected to add a warning here saying the resampling will likely be deprecated in the future.

ConnectedSystems:
Tests fail on CI but pass locally, possibly related #681 .

Merging but I'll do a pass on the tests as well soon.

PR #653, opened 2025-03-16
Add developers setup guide

ConnectedSystems:
Had to reinstall my OS recently so taking the opportunity to properly document setting up a development environment.


tupui:
If we move to uv, which I also recommend, then we can simplify the instructions even more if we use the uv way of working. The `uv pip` interface is just here for compatibility and to help people migrate.

ConnectedSystems:
Closing in favour of #674

Issue #654, opened 2025-04-17
Missing required parameter in the DGSM analysis example doc


singhravi1:
The analysis example for DGSM given [here](https://github.com/SALib/SALib/blob/ac5b42174256b834c61b7512b9e52fe039f81c3a/src/SALib/analyze/dgsm.py#L29) is missing a required parameter X.

Current example is, `Si = dgsm.analyze(problem, Y, print_to_console=False)`  

It should be, `Si = dgsm.analyze(problem, X, Y, print_to_console=False)`

tupui:
Thanks @singhravi1 Docs improvements are welcome if you wanted to contribute.

singhravi1:
Sure! I'm going to raise a pull request soon

PR #655, opened 2025-04-17
fix: dgsm analyze example doc

singhravi1:
This PR fixes an issue #654 related to example for DGSM analyze function.


tupui (approved):
Thanks 😊

PR #661, opened 2025-07-22
Enhance delta.analyze() with step/balanced bootstrap methods and interpretable outputs (fixes #659)

### LINKED ISSUE #659: Substantial (and selective) discrepancy between SALib 1.4.5 and 1.5.1 results in delta.analyze() (Delta Moment-Independent Measure)
I have implemented delta.analyze() for Delta Moment-Independent Measure on my dataset (approx 130 000 measurements). I have noticed some extreme discrepencies between the results from SALib 1.4.5 and SALib 1.5.1, for the same inputs. (See results from the Si returned from delta.analyze())

For easier interpretation, here are the parameters, and the resulting delta score for SALib 1.4.5 and SALib 1.5.1, and I have added the Pearson correlation with the output parameter as well. Will attach the full Si returned by delta.analyze() at the end for reference as well. 

Image

Things to note (indicated in colour):

1. **GREEN** Disproportionately reduced delta scores in v1.5.1. _Note how param05 had the highest delta value in v1.4.5. In v1.5.1, it has less than 50% of the delta score of the highest parameter (param01)_. (and delta of param01 is largely unaffected)
2. **PINK** Delta scores have not changed much. However, because the green parameters have been reduced by so much, they now appear much more important than the green parameters in v1.5.1, whereas, in v1.4.5, they were insignificant compared to the green parameters. 
3. **BLUE** This parameter was one of the most significant in v1.4.5, and is now almost insignificant compared to other parameters in v1.5.1. 

_Relationships and nature of the parameters:_
1. **GREEN** These parameters have some correlation among each other (similar properties and behaviour) and param02-param09 have many zero-values (related to solar radiation over 24-hours)
2. **PINK** These parameters are highly correlated with the output parameter, as well as the param01 (which has the highest delta score).
3. **BLUE** This parameter has many zero-values, and its relationship with the output parameter is suspected to be a _step-like input_. However, the _result does not change when I use only the non-zero values in the input data_, or if I normalize the input set differently (binary (1 or 0) inputs, or logarithmically). I am unable to replicate its high delta score in v1.4.5, no matter how I format the input, or tweak y_resamples in 1.5.1. 

Any help or input would be appreciated, as to which of these results are the most reliable.  

My greatest concern is that the delta values are not changed for all parameters, only a select few (green and blue, 10 out of 22 inputs). This makes input parameters ranking order in terms of importance very different - and greatly affects the interpretation and spread of relative importance. 

The results from v1.4.5 make more sense to me, but I do want to know what the most correct answer is. 

_Note on zero-values:_ The parameters which have the most substantial change from v1.4.5 to v1.5.1 all have quite a few zero-values. However, changing the scaling methods (binary, logarithmic) or resampling using only entries with non-zero values for those parameters, does not affect it to nearly the same extent to achieve similar values to that in v1.4.5.

Full results from delta.analyze() for reference:

Image

Image

Code to generate delta.analyze():
`df = mydf.copy(deep=True) # this df already has all parameters normalized and scaled`
`col_output = 'T'`
`col_input = [col for col in df.columns if col != col_output]`
`df.drop_duplicates(inplace=True)`
`X = df[col_input].values`
`Y = df[col_output].values`
`problem = {
   'num_vars': len(col_input),
   'names': col_input,
   'bounds': [[X[:, i].min(), X[:, i].max()] for i in range(len(col_input))] 
}`
`Si = delta.analyze(problem, X, Y)`
`print(Si)`

ritavdwalt:
### Summary
This PR enhances the existing `delta.analyze()` function in SALib by
- Adding new bootstrap resampling options (`step`, `balanced`)
- Includes raw and resampled delta values for enhanced interpretability and downstream analysis
- Refactored internal logic for improved clarity, flexibility, and result interpretability
- Based on discussion in Issue #659 

### Key changes
1. New bootstrap methods
* `raw` (modified version of the original method, now with Gaussian KDE scaling for improved bandwidth selection)
* `balanced` (bootstrap with uniform sample weights across input distribution)
* `step` (whether the parameter is active >0 (1) or inactive <=0 (0))

2. Extended return dictionary
* Returns additional keys `delta_raw`, `delta_balanced`, `delta_step`, each with corresponding confidence intervals
* Adds a `notes` key that flags potential dataset bias, distribution gaps or unevenness, or bootstrap limitations (eg. a feature always >0 is unsuitable for `step`, so it will not calculate `delta_step`)

3. New user-defined (optional) parameters
* Users can optionally save bootstrapped subsets to a custom path for diagnostics, visual inspection, or reproducibility. (see attached)
* Optional user-specified binning specifications for either number of bins, or custom bin boundaries, with error fallback

4. Testing
* Introduced new unit tests in `test_analyze_delta.py` for each bootstrap method, shape, and type checks, and edge-case validation
* Added internal helpers `create_problemspec()` and `create_df` used only in testing to generate reproducible synthetic data

### Usage example
```
from SALib.analyze import delta

Si = delta.analyze(problem, X, Y)
# New keys:
# Si["delta_raw"], Si["delta_balanced"], Si["delta_step"]
```

### Backward compatibility
1. Fully backward compatible with `SALib==1.4.7`
2. Original outputs `S1`, `S1_conf` are preserved, but to reduce confusion with the modified outputs, `delta` and `delta_conf` are now exposed as `delta_raw` and `delta_raw_conf`

### Testing
**Test summary: (pytest)**
- 204 tests passed
- 2 tests failed _(expected due to KDE scaling change)_
- 1 test marked xfail
- 92 warnings (unrelated to this PR, pre-existing in main SALib test suite)

**Failing tests: (expected)**
1. `tests/test_cli_analyze.py::test_delta`
2. `tests/test_regression.py::test_regression_delta_svm`

Both failures are due to small deviations in the sensitivity indices relative to the expected values with confidence bounds. 
_Reason:_ These tests rely on hard-coded sensitivity values based on the original (unscaled) KDE. The updated version now normalises Gaussian KDEs for proper density scaling, leading to slight shifts in the results. This improves robustness and correctness, but causes expected mismatches with legacy assertions.
_Next step:_ Update expected test outputs for the failing tests, based on the new KDE normalisation (if approved by maintainers)

### Notes
1. These changes enable more nuanced diagnostics and uncertainty quantification of delta indices
4. All additions follow existing `SALib` API and coding conventions (`black`, `flake8`, `ruff` compliant)
5. Improves user result interpretability of delta results in context of both their dataset (variable activity during data capture) and data distribution over its full range. 
6. `pytest` fails two tests (expected), where results are checked against preconfigured values. This is because of the scaling implemented in the gaussian KDE, which now slightly changes resulting delta (`delta_raw`) values. 

### Interpretability
**SUMMARY:**
| Method          | Interpretation question answered | When to Use                         |
|----------------|-----------------------------------------------------------------------------|-------------------------------------|
| `delta_raw`     | How does X (as observed in the dataset) affect Y?                    | 1. When dataset bias is acceptible 2. If you want to see feature influence on **your dataset** |
| `delta_balanced`| How does X across its full range affect Y, regardless of its density in this dataset?                          | When you want to quantify the importance of an input feature on an output feature, without inherent dataset bias.            |
| `delta_step`    | Does presence versus absence of X matter for Y?               | For features where parameters are not always active.    |

1. Earlier version `SALib==1.4.7` answers question:
* "How does a change in feature X **in my dataset** affect output Y distribution **in my dataset**?"
* This answer is dataset-specific, and can be highly biased if input data distribution is more densely populated in a specific region, as the random bootstrap is more likely to use those values in calculations.
* The resulting delta scores from this method should only be compared to one another if the nature of their impact on the output feature is the same, and if the spread of those parameters are similar to one another. This is impractical in most real-world data collection situations. 

2. Modified code: `delta_raw`
* "How does a change in feature X **in my dataset** affect output Y distribution **in my dataset**?"
* Same methodology as in `SALib==1.4.7`, now with gaussian kde scaling implemented

4. Modified code: `delta_balanced`
* "What is the influence of changes in feature X over its distribution, on the distribution of output Y?"
* Examines variation in feature X over its entire spread, regardless of whether the majority of dataset lies in a certain region. Therefore substantially reduces input dataset bias.
* Resulting delta scores are more comparable to one another if the nature of their influence is different to one another, or if one parameter has more variable density in a specific region. 

5. Modified code: `delta_step`
* "What is the impact of this feature when it assumes nonzero (>0) values, regardless of the magnitude of the feature? What is the impact of it just being active, versus inactive?"
* Perfect for parameters which are not always active, when you just want to see the impact of activity.
* Bootstrap is balanced/binned among nonzero values to reduce bias in inputs.
* Resulting delta scores answer the question "how important is whether it rains or not, compared to whether the sun shines or not, regardless of how much it rains, or how much the sun shines?"

### Combination result interpretation: a new and powerful tool
1. Comparing `delta_balanced` and `delta_step` allows distinguishing _if_ a variable matters versus _how much_ its magnitude matters when active.

2. Combining the interpretation of `delta_raw` and `delta_balanced` will provide user insight to how biased their dataset is
* If `delta_raw` and `delta_balanced` differ by more than 0.1, then `notes` will flag it for the parameter. This suggests that the dataset is biased for that input feature, and is densely populated in a specific region. (i.e., the `raw` bootstrap method likely contains values mostly from a specific range. This was the cause for the biased delta scores in Issue #659 )

### Attachments
See attached ZIP folder [`PR_DMIM.zip`](https://github.com/user-attachments/files/21366771/PR_DMIM.zip)  containing example runs on three datasets, with saved bootstraps and delta results. 
Includes
* minimum working example `resultsgen.py`
 which implements the new modified code
* `DATASET_{n}_input.parquet` includes the input datasets used
* `DATASET_{n}_bootstraps.parquet` shows the bootstrap subsets for `raw`, `step` (where applicable), `balanced` methods
* `DATASET_{n}_sensitivity.parquet` shows the sensitivity results, and the `notes` flagged, for all methods.
* `outputplot_{param}.jpg` shows, for each parameter, the plots of the different bootstraps, and resulting delta scores for the new implemented methods.

### DIFF
--- a/src/SALib/analyze/delta.py
+++ b/src/SALib/analyze/delta.py
@@ -254,9 +254,9 @@ def analyze(
                 diff = abs(S["delta_balanced"][i] - S["delta_raw"][i])
                 if diff > delta_warn_threshold:
                     note += f"Raw delta differs from balanced by {diff:.2f}; "
-                    # warnings.warn(
-                    #     f"[{name}] Potential Bias Notice: Raw delta score differs from balanced delta by {diff:.2f}. Potential dataset bias, take care in interpretation."
-                    # )
+                    warnings.warn(
+                        f"[{name}] Potential Bias Notice: Raw delta score differs from balanced delta by {diff:.2f}. Potential dataset bias, take care in interpretation."
+                    )
 
             if "sobol" in methods:
                 ind = np.random.randint(Y.size, size=y_resamples)
@@ -318,7 +318,7 @@ def check_specified_bininfo(bininfo, Xmin, Xmax, paramname):
         bin_edges = np.array(bininfo)
         if not np.issubdtype(bin_edges.dtype, np.number):
             warnings.warn(
-                f"[{paramname}] Input Error: Bin value error: Custom bin list only contain numeric values. Resorting to default binning ({defaultbins} bins)."
+                f"[{paramname}] Input Error: Bin value error - Custom bin list only contain numeric values. Resorting to default binning ({defaultbins} bins)."
             )
             return (
                 np.linspace(Xmin, Xmax, defaultbins + 1),
@@ -326,7 +326,7 @@ def check_specified_bininfo(bininfo, Xmin, Xmax, paramname):
             )
         elif np.any(bin_edges < Xmin) or np.any(bin_edges > Xmax):
             warnings.warn(
-                f"[{paramname}] Input Error: Bin boundary error: Custom bin boundaries must be within the X range. Resorting to default binning ({defaultbins} bins)."
+                f"[{paramname}] Input Error: Bin boundary error - Custom bin boundaries must be within the X range. Resorting to default binning ({defaultbins} bins)."
             )
             return (
                 np.linspace(Xmin, Xmax, defaultbins + 1),
@@ -334,7 +334,7 @@ def check_specified_bininfo(bininfo, Xmin, Xmax, paramname):
             )
         elif not np.all(np.diff(bin_edges) > 0):
             warnings.warn(
-                f"[{paramname}] Input Error: Bin edge error: Custom bin edges must be ascending. Resorting to default binning ({defaultbins} bins)."
+                f"[{paramname}] Input Error: Bin edge error - Custom bin edges must be ascending. Resorting to default binning ({defaultbins} bins)."
             )
             return (
                 np.linspace(Xmin, Xmax, defaultbins + 1),
@@ -343,7 +343,7 @@ def check_specified_bininfo(bininfo, Xmin, Xmax, paramname):
         return np.concatenate(([Xmin - 1e-9], bin_edges, [Xmax + 1e-9])), ""
     else:
         warnings.warn(
-            f"[{paramname}] Input Error: Invalid custom bin dtype: Must be int, list or None. Resorting to default binning ({defaultbins} bins)."
+            f"[{paramname}] Input Error: Invalid custom bin dtype - Must be int, list or None. Resorting to default binning ({defaultbins} bins)."
         )
         return (
             np.linspace(Xmin, Xmax, defaultbins + 1),
@@ -500,6 +500,12 @@ def obtain_resampled_subset(
                 f"[{paramname}][{mode}] Only one bin remains. Revisit whether processing method is suitable for this parameter, or increase dataset size or spread. Highly biased input.",
                 f"[{mode}] Single valid bin. Insufficient spread in feature, highly biased input; ",
             )
+
+        if n_bins != len(init_indices):
+            if warn_print:
+                warnings.warn(
+                    f"[{paramname}][{mode}] Bin Merge Notice: Final no. bins is {len(final_indices)}. Min samples per bin: {min_class_size}"
+                )
 
         if n_per_bin < min_class_size:
             raise ValueError(
--- a/tests/test_analyze_delta.py
+++ b/tests/test_analyze_delta.py
@@ -1,27 +1,22 @@
 from SALib.analyze import delta
+from SALib.util import handle_seed
 import numpy as np
 import pandas as pd
 import pytest
 
 
-def create_base_dataframe(seed=None):
-    if seed is None:
-        seed = 42
-    np.random.seed(seed)
+def create_base_dataframe(seed=42):
+    rng = handle_seed(seed)
 
     n = 15000
-    col_a = np.linspace(10, 80, n, endpoint=False) + np.random.uniform(
-        0, 70 / n, size=n
-    )
-    col_b = np.linspace(-20, 50, n, endpoint=False) + np.random.uniform(
-        0, 70 / n, size=n
-    )
-    np.random.shuffle(col_a)
-    np.random.shuffle(col_b)
+    col_a = np.linspace(10, 80, n, endpoint=False) + rng.uniform(0, 70 / n, size=n)
+    col_b = np.linspace(-20, 50, n, endpoint=False) + rng.uniform(0, 70 / n, size=n)
+    rng.shuffle(col_a)
+    rng.shuffle(col_b)
     df = pd.DataFrame(
         data={
             "output_col": np.clip(
-                np.random.normal(loc=22, scale=1.3, size=n), 17, 32
+                rng.normal(loc=22, scale=1.3, size=n), 17, 32
             ),  # Normal distribution centered at 22 with small std dev
             "column_a": col_a,
             "column_b": col_b,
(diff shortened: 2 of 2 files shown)


21574081:
I also considered that, but because we require `num_resamples` number of 'random' subsets from the data, it doesn't seem feasible to do the balancing before calling `delta.analyze()`, just because we want to have access to all available samples when doing the bias-reduced delta. 

If the user implements the 'balancing' before calling `delta.analyze()`, then the library will do bias-reduced delta on `num_resamples` variations of the balanced input, but all other samples that were not selected by balancing will subsequently be ignored. 

Also, the issue is that, currently, the library is being used without any data preparation or balancing, and the highly biased results end up being published blindly. The intention behind my application is to at least make users consider the different ways how different inputs can impact the output. 

Even if it isn't perfect, it should provide more insight than the current implementation, and, for now, improve interpretability. 


One solution could be to implement a new function in `delta.py` which is seen as a pre-processing function to be called before `analyze()`, which up-samples under-represented bins to create a more balanced (but much larger) dataset. In doing that, all data samples are included in the inputs X, Y, but this will have a substantial trade-off with computational cost.

ConnectedSystems:
Hi @21574081 apologies for the long delay, I'm only just finding the time to have a look at SALib again.

There are two key aspects to the discussion above which I think can be handled through iterative improvement.

First, the question of bootstrapping. I, too, would like to move to a more standardized approach. Currently different methods apply different bootstrapping approaches. Whether it is for a specific reason for a given approach or not, and how much alignment between approaches can be achieved requires careful assessment. So this is a longer term improvement in my view.
I am more than okay with accepting your bespoke bootstrapping approach specifically for `delta`, at least for now.

Second, balanced sampling, as @tupui says, is a whole can of worms in itself. We can't address everything and it would be folly to try - there's no such thing as a perfect method or approach, nor a perfect software package. The best we can do, in my view, is to do our best to document these limitations and clearly outline what is expected of the user, in terms of data provision. The iterative improvement we can do is identify where/when it is easy to detect cases and perhaps throw a warning. Otherwise it's down to documentation and raising awareness amongst users/researchers/practitioners.

Lastly, there is now #662 which may affect this PR. I'll be taking a look shortly and judging which should be merged and what adjustments are needed to align changes. So please don't be offended if the later PR gets handled first - I know you've been waiting for quite some time on this.

ConnectedSystems:
Incidentally, could you please allow maintainers to push commits to your PR please? Thanks.

21574081:
Hi @ConnectedSystems thanks for getting back to me!

I saw PR #662 and agree that it makes sense to address it preceding this PR as it affects this one as well.

I agree with you that, at least for now, iterative enhancements for improved result integrity is the way to go, with longer-term approaches toward more robust methodology. 

I also agree that the fine-tuning and assurance of result integrity is highly contextual and should be the responsibility of the researcher/user/practitioner. However, unfortunately, many users may not realise that this should be done, or have the knowledge of how to statistically correctly implement these methods if their relationship of influence toward the output is not the same. 

I do think, for now, the iterative enhancement of introducing different metrics for the `delta` method will at least flag to users or encourage users to go look and investigate _why_ there are different metrics and _which_ is most appropriate to them. This may at the very least encourage users to investigate the methods before blindly applying them and can reduce false results.

ritavdwalt:
@ConnectedSystems Checking in to see whether any actions are required on my end to assist in the integration?

ConnectedSystems:
@ritavdwalt thanks for your patience, I got heavily distracted with work hence the delay.

Any remaining issues can be addressed in a future PR so have merged now. Thank you for your contribution!

Issue #634, opened 2024-09-05
Standardized system for parameter management across Scientific Python libraries


EwoutH:
_This isn't an issue, but SALib doesn't have GitHub [Discussions](https://docs.github.com/en/discussions) enabled. Hope it's okay that I posted here._

Over at Mesa we’re working on a system for managing model/agent default values and ranges, and I believe this could be beneficial beyond Mesa. We’re exploring how to create a standardized way to define parameter spaces, including default values, ranges, and distributions, to be used by any internal and potentially external components.

The idea is to develop a `ParameterSpec` structure that can be used for continuous, discrete, ordered, and categorical data, with flexibility for specifying distributions, sampling strategies, and default values. Then each software component (samplers, visualisation, batch runners) can extract the pieces of information it needs. By defining everything in one place, it makes it easier to integrate different components and much less error prone.

Since we also did something similar in the EMAworkbench (see [parameters](https://emaworkbench.readthedocs.io/en/latest/ema_documentation/em_framework/parameters.html)), we where thinking such a spec might be beneficial for more simulation and other scientific libraries. It might have potential for a [SPEC](https://scientific-python.org/specs/).

We’ve started [discussions here](https://github.com/projectmesa/mesa/issues/2268), and we’d love to hear your thoughts and see if there’s potential for a broader collaboration. If there’s much interest, we can start a centralized discussion with the Scientific Python community.

- https://github.com/projectmesa/mesa/issues/2268

tupui:
Thanks for including us in the discussion. I would need to dive deeper but sounds like a good idea to try to come up with a standard. This is something a lot of libraries are re-inventing all the time. I am of course all for a SPEC (I can help on that front) 😃

Is there some meetings we can setup for that? Also we could discuss on the Scientific Python discord directly. We have a SALib channel where that could take place for instance.

Also note that in SciPy we are shipping soon a new infrastructure for distributions. So it would be great to consider this. The SciPy spec does not deal with things like Copulas and this is right now a thing in statsmodels so also to consider I think (I have my reservations on statsmodels but nevertheless.)

EwoutH:
Thanks for your enthusiastic reply! Right now this is nothing more than an idea, which I was curious about. All discussion so far has taken place in the issue linked above.

[...]

PR #662, opened 2025-09-08
Shift to using numpy.random.Generator and harmonize signature of seed across all sample functions

quaquel:
This PR updates all sample functions to accept an `integer`, `None`, or a `numpy.random.Generator` instance as an argument for seed. It also ensures that this generator is used throughout the code to ensure reproducibility of the results. 

I have kept `seed` as the name of the keyword argument. For [spec-7](https://scientific-python.org/specs/spec-0007/) compliance, this should be renamed to `rng`. However, changing this would be backward incompatible. Of course, it would be easy to add it in here and, for the time being, accept either seed or rng and raise a deprecation warning on seed. If that is desirable, this is easy to add. 

See #634 for some background


EwoutH:
Without having reviewed the code line-by-line, this looks like a proper step towards standardizing seed passing and SPEC 7 compliance.

quaquel:
The unit tests fail before the actual testing starts. 
* 3.9 errors on setting up conda with `Did not find Mambaforge-Linux-x86_64.sh latest in cache`
* 3.12 errors on setting up conda with a 404 error.

Linter caught a couple of small issues with morris, which are now fixed. The other linter issue is that seed/rng is present in the signature of full factorial and finite difference, but not used internally by either method. I cannot really fix this. The issue was there before this PR already.

ConnectedSystems:
Thanks @quaquel 

>Linter caught a couple of small issues with morris, which are now fixed. The other linter issue is that seed/rng is present in the signature of full factorial and finite difference, but not used internally by either method. I cannot really fix this. The issue was there before this PR already.

SALib methods have to define a seed/rng argument to maintain compatibility with the CLI.
There are ways to handle methods that do not require a seed value at all of course, but it's out of scope here I think.

Could anyone point me to a draft/interim SPEC7 document? As far as I can tell it's [not yet finalised](https://hackmd.io/@scientific-python/bof2024?utm_source=preview-mode&utm_medium=rec) but I want to see how far we are in alignment with current discussions at least.

ConnectedSystems:
@tupui 

Did scipy change the interface for `scipy.stats.uniform` recently?

Looks like we had been running `scipy.stats.uniform(loc=sample * d, scale=sample * d, size=num_groups, seed=<some value>)` for years, but now it doesn't work (seed argument not supported).

Looks like the correct usage is now `scipy.stats.uniform.rvs(loc=sample * d, scale=sample * d, size=num_groups)`

But now all tests that check results within a certain tolerance (for LHS sampling) appear to fail?

tupui:
Mmm there was some changes to use rng instead of seed. I am surprised it's breaking without warning though.

But in general for basic distributions there is the new API which is recommended because better in every way.

tupui:
The SPEC-7 is here otherwise https://scientific-python.org/specs/spec-0007/

quaquel:
First, thanks for the commits and minor modifications.

> SALib methods have to define a seed/rng argument to maintain compatibility with the CLI.
There are ways to handle methods that do not require a seed value at all of course, but it's out of scope here I think.

Yes, let's keep this out of scope for this PR. And indeed it is relatively easy to fix. I would suggest checking the validity of the CLI arguments for each method, rather than having a common interface with arguments that for certain methods don't matter.


> Could anyone point me to a draft/interim SPEC7 document? As far as I can tell it's [not yet finalised](https://hackmd.io/@scientific-python/bof2024?utm_source=preview-mode&utm_medium=rec) but I want to see how far we are in alignment with current discussions at least.

SPEC-7 is endorsed by `ipython`, `networkx`, `numpy`, `scikit-image`, and `scipy`, so I guess it has been finalized. The only discrepancy between this PR and spec-7 is that this PR still uses `seed` instead of `rng` as recommended by SPEC-7. Apart from that, everything is compliant. The simplest ways to make this PR fully compliant, without breaking backward compatibility, would be to:
1. Accept either `seed` or `rng`. 
2. Raise a `DeprecationWarning` on `seed` and set `rng = seed`. 
3. Raise a `ValueError` if both `seed` and `rng` are passed.

It would be trivial to embed this logic in a function decorator and just decorate all `sample` functions with it. Once the use of seed is fully deprecated, just remove the decorator and you're done.

> But now all tests that check results within a certain tolerance (for LHS sampling) appear to fail?

Not sure what's happening on the tests. I ran all tests locally before creating the PR, and everything was fine on my end.

quaquel:
I have done some further cleaning and reverted some changes in the `analyze` functions. At present, quite a few of the `analyze` functions also use `np.random.seed`. While fixing failing unit tests, I started to change this as well, but that made this PR too big and unwieldy. 

I now have the following failing tests


```console

FAILED tests/sample/morris/test_morris.py::TestGroupSampleGeneration::test_generate_x_star - AssertionError: 
FAILED tests/sample/test_latin.py::TestLatinSample::test_latin_sample_trivial - AssertionError: 
FAILED tests/sample/test_latin.py::TestLatinSample::test_latin_sample_trivial_group - AssertionError: 
FAILED tests/sample/test_latin.py::TestLatinSample::test_latin_sample_one_group - AssertionError: 
FAILED tests/sample/test_latin.py::TestLatinSample::test_latin_sample_two_groups - AssertionError: 
FAILED tests/test_cli_analyze.py::test_fast - AssertionError: Unexpected FAST results.
FAILED tests/test_cli_analyze.py::test_morris - AssertionError: Results did not match expected values:
FAILED tests/test_cli_analyze.py::test_morris_scaled - AssertionError: Results did not match expected values:
FAILED tests/test_regression.py::TestMorris::test_regression_morris_vanilla - AssertionError: 
FAILED tests/test_regression.py::TestMorris::test_regression_morris_scaled - AssertionError: 
FAILED tests/test_regression.py::TestMorris::test_regression_morris_groups - AssertionError: 
FAILED tests/test_regression.py::TestMorris::test_regression_morris_groups_brute_optim - AssertionError: 
FAILED tests/test_regression.py::TestMorris::test_regression_morris_optimal - AssertionError: 
FAILED tests/test_util.py::test_nonuniform_scale_samples_truncnorm - AssertionError: 
```

As far as I can tell, having checked each of these tests, these assertions all fail because they test for specific return values. However, since the algorithm for the random number generation has changed, the number evidently also will change, so it's not surprising that these tests are failing. 

Please let me know how you want me to continue with this PR.

PR #674, opened 2026-03-06
Updated development setup and release guide

ConnectedSystems:
Walk newcomers through setting up development environment using `uv`.

Disclaimer:

The development setup guide was 95% written by me. I used Claude to indicate `pip` equivalent commands so those unfamiliar with `uv` have a conceptual anchor.

The release guide was 80% AI generated (with Claude). I intend to follow it step by step with the next release and adjust where changes/clarifications are necessary.

@tupui appreciate a quick lookover for anything that looks completely wrong.

PR #675, opened 2026-04-01
Add bounds check for log-uniform distributions

### LINKED ISSUE #683: Bounds checks for log-normal distributions
`logunif` bounds are not validated in `_nonuniform_scale_samples`. Every other distribution branch (`unif`, `norm`, `truncnorm`, `lognorm`, `weibull`) checks its parameters and raises a `ValueError` on invalid input. The `logunif` branch skips this and passes bounds straight to `scipy.stats.loguniform.ppf`:

https://github.com/SALib/SALib/blob/main/src/SALib/util/__init__.py#L210-L211

```python
elif dists[i] == "logunif":
    conv_params[:, i] = sp.stats.loguniform.ppf(params[:, i], a=b1, b=b2)
```

When the bounds are invalid for a log-uniform distribution (lower bound `<= 0`, since `log(x <= 0)` is undefined, or lower bound `>=` upper bound), scipy silently returns `NaN` instead of raising. SALib then propagates these `NaN` values into the generated sample set with no warning or error.

## Reproduction

```python
import numpy as np
from SALib.util import scale_samples

rng = np.random.default_rng(42)

# Case 1: lower bound <= 0
problem_neg_lower = {
    "num_vars": 1,
    "names": ["x1"],
    "bounds": [[-1, 10]],
    "dists": ["logunif"],
}
params = rng.uniform(size=(5, 1))
scaled = scale_samples(params.copy(), problem_neg_lower)
print(scaled.ravel())
# [nan nan nan nan nan]

# Case 2: lower bound >= upper bound
problem_bad_order = {
    "num_vars": 1,
    "names": ["x1"],
    "bounds": [[10, 1]],
    "dists": ["logunif"],
}
params = rng.uniform(size=(5, 1))
scaled = scale_samples(params.copy(), problem_bad_order)
print(scaled.ravel())
# [nan nan nan nan nan]
```

## Expected behavior

SALib should validate `logunif` bounds and raise a `ValueError` (lower bound must be `> 0` and less than upper bound), consistent with the existing checks for `unif`, `norm`, `truncnorm`, and `weibull`.

## Actual behavior

Invalid bounds are silently converted to `NaN`, which then propagate undetected through sampling and analysis.

--- 

## Use of AI disclaimer

I wrote the fix myself but used Claude to provide an overview (above) for the requested issue in #675.

ConnectedSystems:
Previously, SALib was silently filling with NaN values.

Blocked by #674 (should be updated from main branch after #674 is merged)

### DIFF
--- a/docs/developers_guide.md
+++ b/docs/developers_guide.md
@@ -1,23 +1,98 @@
 # Developers Guide
 
+## Project setup
+
+For new developers: `git` must be installed.
+
+The project and its dependencies are managed via `hatch` and `uv`.
+
+While it is possible to maintain a local version of this project using venv/pip/mamba/conda,
+to simplify this guide, `uv` and `hatch` is the focus.
+
+Install `uv` following instructions from [here](https://docs.astral.sh/uv/getting-started/installation/).
+
+```bash
+$ uv venv                    # equivalent: python -m venv .venv
+$ .venv/Scripts/activate
+```
+
+On Windows, the commands above should work for those using PowerShell.
+
+Otherwise, run the below for classic command prompt:
+
+```bash
+> uv venv
+> .venv/Scripts/activate.bat
+```
+
+Then install `hatch` as a tool:
+
+```bash
+$ uv tool install hatch      # equivalent: pip install --user hatch (but isolated)
+```
+
+Verify installation after restarting terminal with:
+
+```bash
+$ hatch --version
+```
+
+Build SALib with:
+
+```bash
+$ hatch build
+```
+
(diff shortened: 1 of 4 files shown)


willu47 (commented):
Can you link to an issue where this problem is described, or add a failing test which demonstrates the erroneous behaviour, which is subsequently fixed by your contribution?

ConnectedSystems:
Hi @willu47 I'll be merging this one in soon if that's okay with you. 
It's really a two `if statement` fix to catch an edge case.

PR #677, opened 2026-04-11
fix: wrap p_imap result with list() to prevent TypeError

### LINKED ISSUE #676: TypeError with p_tqdm available and multiple cores
In SALib/util/problem.py, line 238:
```python
  2         if ptqdm_available:
  1             # Display progress bar if available
238             res = p_imap(tmp_f, chunks, num_cpus=nprocs)
  1         else:
  2             with Pool(nprocs) as pool:
  3                 res = list(pool.imap(tmp_f, chunks))
```
should be:
```python
238             res = list(p_imap(tmp_f, chunks, num_cpus=nprocs))
```

Otherwise, using the last example from [SALib wrappers documentation](https://salib.readthedocs.io/en/latest/user_guide/wrappers.html) gives TypeError **if p_tqdm is available** and multiple cores are use in `evaluate`.

The complete example to run:
```python
# p_tqdm must be installed to see the bug!

import numpy as np

def linear(a: float, b: float, x: float) -> float:
    return a + b * x


def wrapped_linear(X: np.ndarray, func=linear) -> np.ndarray:
    N, D = X.shape
    results = np.empty(N)
    for i in range(N):
        a, b, x = X[i, :]
        results[i] = func(a, b, x)

    return results


from SALib import ProblemSpec

sp = ProblemSpec({
    'names': ['a', 'b', 'x'],
    'bounds': [
        [-1, 0],
        [-1, 0],
        [-1, 1],
    ],
})

(
    sp.sample_sobol(2**6)
    .evaluate(wrapped_linear, nprocs=2)
    .analyze_sobol(print_to_console=True)
)
```

thakoreh:
## Summary

Fixes #676

When `p_tqdm` is available and multiple cores are used for parallel evaluation, `p_imap` returns a generator that cannot be properly consumed by `_collect_results()`.

## Change

Wrap the `p_imap()` call with `list()` in `evaluate_parallel()` (line 238 of `src/SALib/util/problem.py`).

```python
# Before
res = p_imap(tmp_f, chunks, num_cpus=nprocs)

# After
res = list(p_imap(tmp_f, chunks, num_cpus=nprocs))
```

## Test Plan

Added `tests/test_ptqdm_fix.py` with 3 tests:
- `test_evaluate_with_ptqdm_returns_consumable_results` — mocks p_imap as a generator, verifies results are consumed correctly
- `test_evaluate_without_ptqdm_still_works` — regression test for non-p_tqdm path
- `test_evaluate_single_core_no_ptqdm` — single-core fallback

All tests pass.

## Verification

```
$ python -m pytest tests/test_ptqdm_fix.py -v
3 passed
```

### DIFF
--- a/src/SALib/util/problem.py
+++ b/src/SALib/util/problem.py
@@ -235,7 +235,7 @@ class ProblemSpec(dict):
 
         if ptqdm_available:
             # Display progress bar if available
-            res = p_imap(tmp_f, chunks, num_cpus=nprocs)
+            res = list(p_imap(tmp_f, chunks, num_cpus=nprocs))
         else:
             with Pool(nprocs) as pool:
                 res = list(pool.imap(tmp_f, chunks))
--- a/tests/test_ptqdm_fix.py
+++ b/tests/test_ptqdm_fix.py
@@ -1,0 +1,75 @@
+"""Test that evaluate() returns consumable results when p_tqdm is available.
+
+Regression test for https://github.com/SALib/SALib/issues/676
+Without the fix, p_imap returns a generator instead of a list, causing
+TypeError when _collect_results() tries to consume it.
+"""
+
+import numpy as np
+import pytest
+from unittest import mock
+
+from SALib import ProblemSpec
+from SALib.test_functions import Ishigami
+
+
+def _make_spec():
+    """Create a basic ProblemSpec with samples."""
+    sp = ProblemSpec(
+        {
+            "names": ["x1", "x2", "x3"],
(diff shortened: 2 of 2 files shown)


tupui:
That's a first I think. Before doing anything, I think we need a AI policy. I suggest taking the one from SymPy and LLVM. What we need is some official disclosure on all PRs. Otherwise we are at risk of breaking our licensing and copyright guarantees. And of course quality.

ConnectedSystems:
I can confirm the reported issue is real and the fix is valid as far as I can tell.

@tupui lets discuss the disclosure question in a separate thread

tupui:
Oh totally I just meant to say we need to think about AI, not that this was incorrect 😃

PR #686, opened 2026-09-03
Add Goda Shapley effects method

### LINKED ISSUE #682: Implement method of Goda for Shapley effects estimation
[Goda's method](https://doi.org/10.1016/j.ress.2021.107702) provides a relatively straightforward approach to obtaining Shapley effect values.

This is already implemented [in Julia](https://github.com/Zapiano/SAShE.jl) by my colleague @Zapiano and a port to SALib is doable.

I also contributed (in the above) changes to support given-data contexts for user accessibility.

DresdenGman:
# Add Goda Shapley effects method

Closes #682.

## Summary

This adds a Shapley-effects implementation based on the random-permutation estimator described by Goda (2021). The implementation follows SALib's existing sample/evaluate/analyze workflow:

- `SALib.sample.shapley` generates the paired hybrid samples required by the estimator using `N * (D + 1)` model evaluations.
- `SALib.analyze.shapley` returns effects in the model-output variance scale, confidence intervals derived from the estimator's sample variance, and a `normalized` view whose effects sum to one.
- The method is available through the command-line interface and `ProblemSpec` dynamic methods.
- The documentation includes Python, POSIX shell, and Windows batch examples.

The initial scope supports independent, ungrouped inputs. A user-provided conditional-sampling callback for dependent inputs would require an API that can participate before external model evaluation; I have left that extension out of this PR so it can be designed separately without coupling sampling to model execution.

## Validation

- Ishigami regression against established Shapley-effect reference values
- Exact paired-difference variance identity on an additive linear model
- Input-shape, constant-output, grouping, and invalid-configuration checks
- CLI sample-to-analyze round-trip
- 25 focused Shapley tests passing locally
- Full test suite: 232 passed, 1 expected xfail; one pre-existing stochastic delta regression failed once and passed when rerun in isolation


DresdenGman:
The failed CI job is isolated to `tests/test_regression.py::test_regression_delta_svm`, a stochastic delta-regression test outside this PR's Shapley changes. The branch is current with `main`, and the 25 focused Shapley tests (including the CLI round trip) pass locally.

I attempted to rerun the failed workflow, but GitHub requires repository-admin permission. If convenient, could a maintainer rerun the pytest workflow? I will address any failure that reproduces in the Shapley scope.

ConnectedSystems:
Many thanks for this implementation @DresdenGman 

I've added the performance improvement I mentioned which vectorises the nested for-loops. Looping in Python is comparatively slow so we see a 15-50x improvement:

| N | D | old (s) | new (s) | speedup |
|---|---|---|---|---|
| 1,000 | 3 | 0.0128 | 0.0008 | 15.2x |
| 20,000 | 3 | 0.2593 | 0.0066 | 39.4x |
| 10,000 | 10 | 0.4684 | 0.0089 | 52.8x |
| 50,000 | 8 | 1.6874 | 0.0374 | 45.1x |
| 5,000 | 50 | 1.0554 | 0.0518 | 20.4x |

Timings taken with the help of AI.

It does introduce increased memory use but it is only a concern for very large number of samples and dimensions.
For usual problem sizes it won't be an issue but I have an idea to batch the analyses that will resolve this as well.

| N | D | old peak | new peak | ratio |
|---|---|---|---|---|
| 20,000 | 3 | 3.05 MB | 5.29 MB | 1.7x |
| 5,000 | 50 | 106.05 MB | 126.87 MB | 1.2x |
| **1,000** | **500** | **4.0 MB** | **270.6 MB** | **67x** |

PR #694, opened 2026-09-21
Add retained resamples to Morris analysis

### LINKED ISSUE #579: Enhancing Morris Method: Returning Sensitivity Indices for Resampling Analysis
Hello, 
To facilitate the analysis of resampling results in Morris method, it would be beneficial to add the functionality to return sensitivity indices for each resample. Similar to the "Keep resamples" feature in Sobol' method that allows storing and returning the first-order and total-effect sensitivity indices for each resample, a similar update would enhance the Morris method.

By adding this capability, researchers can access and analyze the sensitivity indices obtained from each individual resample. This would enable further investigation of the variation between resamples through external experiments and analyses. For instance, it would be possible to examine the differences in ranking results across all resamples.

Implementing this enhancement in Morris method would provide users with more comprehensive insights into the sensitivity analysis process and allow for a more thorough evaluation of the impact 

of different resampling iterations. It would be valuable for researchers seeking to gain a deeper understanding of the robustness and reliability of their sensitivity analysis results.

If such an update is implemented, it would enhance the functionality and flexibility of the Morris method, providing users with the ability to analyze and interpret the sensitivity indices obtained from each resample.

DresdenGman:
## Summary

- add an opt-in `keep_resamples` argument to Morris analysis, returning the bootstrap `mu_star` estimates as `mu_star_conf_all`
- use one trajectory-resample index matrix across parameters so each retained row is a coherent bootstrap replicate for ranking and robustness analysis
- preserve DataFrame, plotting, and `print_to_console` behavior by excluding the two-dimensional resample matrix from tabular conversion

## Details

The retained matrix has shape `(num_resamples, number_of_groups)`. Without groups, its columns follow the problem parameter order. With groups, each bootstrap replicate is aggregated using the same group mean as the other Morris `mu_star` outputs.

The storage flag does not change confidence results: both paths use the same shared trajectory draws for a given seed. Morris previously generated independent trajectory indices inside the parameter loop. That was sufficient for marginal confidence widths, but it could not produce a coherent per-resample ranking across parameters. Moving the draw outside the loop preserves the estimator while aligning each bootstrap replicate. As a result, exact seeded confidence values can change from earlier versions; the CLI regression fixtures are updated to record that intentional numerical change.

For grouped analyses, the existing `mu_star_conf` definition remains unchanged: it is the mean of the parameter-level confidence widths. It therefore is not generally identical to a confidence width recomputed from an already-grouped `mu_star_conf_all` column.

Closes #579.

## Validation

```text
245 passed, 1 xfailed
```

```bash
pytest -q
ruff check src/SALib/analyze/morris.py tests/test_analyze_morris.py
black --check src/SALib/analyze/morris.py tests/test_analyze_morris.py
git diff --check
```

Regression coverage verifies the opt-in/default contract, confidence invariance, shared trajectory alignment, reconstruction from retained estimates, grouped aggregation, default key compatibility, and DataFrame/console conversion.


willu47:
@kerained we have received a submission addressing your suggestion for SALib. Would you be interested in assessing if it meets your needs?

